Issue #15, opened 2015-10-06
Carthage Build Error


cocoahero:
Building the framework using `carthage update --platform ios` produces an error of:

```
error: cannot parse the debug map for ".../Carthage/Build/iOS/Buy.framework/Buy": The file was not recognized as a valid object file
```

Xcode 7.0.1, Carthage 0.9.2

runmad:
Feel to try out https://github.com/Shopify/mobile-buy-sdk-ios/releases/tag/1.2.1 which has a bunch of fixes

PR #17, opened 2015-10-06
Carthage Build Fixes

cocoahero:
This patch fixes the library's compatibility with Carthage (shopify/mobile-buy-sdk-ios#15). Notable changes are:

-- Removed custom Mach-O type (now proper dynamic framework)
-- Removed custom `.modulemap`. This is generated automatically by Carthage and Cocoapods.

This may affect the "Universal Framework" target / build script. Not sure if that target is even necessary anymore given that both Carthage and Cocoapods have built in mechanisms for handling unused / unsupported architecture slices.


davidmuzi:
Thanks for the fix!  We don't want to modify the existing target for the static framework, as we rely on it for our fat builds - which we will continue to build for releases.

We are working on a fix by adding another target for the dynamic framework which will work with Carthage.

davidmuzi:
Fix incoming which addressed Carthage and Cocoapods.  Closing this PR.

Issue #16, opened 2015-10-06
Add nullability and generics attributes to all public API


calebd:

calebd:
Because Swift.

cocoahero:
WIP branch: https://github.com/hodinkee/mobile-buy-sdk-ios/tree/generics-nullability

ollieatkinson:
:+1:

calebd:
Bumping this issue.

bgulanowski:
Should be solved by version 2.0. Please see develop branch.

PR #20, opened 2015-10-07
Generics / Nullability Annotations

cocoahero:
This patch adds generics and nullability annotations to most obvious public APIs. (shopify/mobile-buy-sdk-ios#16)

There are several locations that it does not touch, such as model JSON deserialization and their respective properties. It would probably be best to make a design decision in terms of whether model properties should be considered optional or not and the deserialization system should take that into account.


davidmuzi:
This change looks good to merge.

We'll need to add more annotations to the APIs, and down to the deserialization, as noted.

PR #23, opened 2015-10-08
Dynamic framework, Cocoapods fixes, Carthage support, readmes

### LINKED ISSUE #15: Carthage Build Error

### LINKED ISSUE #21: Cannot use @import Buy; from Cocoapods
If using Cocoapods, I cannot use @import Buy; at all
Error message is: Module 'Buy' not found

Do you guys have any idea?

runmad:
Fix #15
Fix #21 

### What this does

* Adds a Dynamic Framework target
* Sample apps updated to embed the dynamic framework
* The workspace has been removed, please open the sample apps and the SDK project files individually
* READMEs for sample apps moved from Shopify docs to their repos.
* Moved `Buy-Bridging-Header` from the classes folder to the Swift sample app
* Fixed Cocoapods support
* Update to version `1.2.1`


.gitignore
```diff
@@ -9,14 +16,23 @@ Ibuild/*
...
 *-INTERNAL.*
+
+## Obj-C/Swift specific
+*.hmap
+*.ipa
+
+# Carthage
+#
+# Add this line if you want to avoid checking in source code from Carthage dependencies.
+# Carthage/Checkouts
```
davidmuzi: :+1:

README.md
```diff
@@ -42,6 +45,12 @@ pod "Mobile-Buy-SDK"
 
 Then run `pod install`
 
+Then run `pod install`
```
davidmuzi: dedupe

README.md
```diff
@@ -92,9 +101,11 @@ The workspace includes the Mobile Buy SDK project and the sample app projects. T
...
+* **Buy**: This is the Mobile Buy SDK static framework. This build is based on the current build configuration. To build a universal framework that can run on a device and on the Simulator and to be included in your app, please refer to the `Universal Framework` target below
 
-* **Universal Framework**: This builds the framework using `build_universal.sh` script in the `Universal Framework` target and copies the built framework in the `/Mobile Buy SDK Sample Apps` folder. Build this target if you have made any changes to the framework that you want to test with the sample apps as the sample apps do not build the framework directly but embed the already built framework
+* **Buy Dynamic**: This is the Mobile Buy SDK dynamic framework. Please refer to the installation section above.
+
+* **Universal Framework**: This builds a static framework using `build_universal.sh` script in the `Universal Framework` target and copies the built framework in the `/Mobile Buy SDK Sample Apps` folder. Build this target if you have made any changes to the framework that you want to test with the sample apps as the sample apps do not build the framework directly but embed the already built framework
```
davidmuzi: since you modifying, can you mention this is a fat binary including arm and i386 slices

davidmuzi:
Do we still need a workspace?

runmad:
I left the workspace there since the test depend on it. We don't necessarily need it. I can update the test.

davidmuzi:
Looks good :ship:

cocoahero:
This does not completely fix the carthage problem. If you try and run `carthage update`, you will see it attempt to build both framework targets, the faux framework, and the dynamic framework.

```
*** Building scheme "Buy Dynamic" in Mobile Buy SDK.xcodeproj
*** Building scheme "Buy" in Mobile Buy SDK.xcodeproj```

The build fails for the same reason as before because of the broken framework target.

runmad:
@cocoahero It seems Carthage is kind of random when it comes to which scheme/target it builds first (and what gets built). I have re-structured what is included in the schemes, PR upcoming: https://github.com/Shopify/mobile-buy-sdk-ios/pull/30

cocoahero:
@runmad Carthage will only build schemes that are shared, as per their documentation:

https://github.com/Carthage/Carthage#share-your-xcode-schemes

runmad:
It's shared: https://github.com/Shopify/mobile-buy-sdk-ios/blob/runmad.fix-frameworks-carthage/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK.xcodeproj/xcshareddata/xcschemes/Buy.xcscheme

Carthage was building the previously shared dynamic and static schemes. The PR removes the static scheme since it's being built by the script scheme.

calebd:
So much :-1:. Why is this still a framework and not a `.a`. Static frameworks break so many things. There is no reason to distribute code this way.

calebd:
(For the record. I don't want a static library at all. Nobody will use it given both Carthage and CodoaPods as alternatives. It's not worth the trouble.)

davidmuzi:
@calebd We understand you may not want the static library, but we have other developers who depend on it.  The framework allows us to package up the headers and assets as well.

We may consider dropping static in a next major release, but not at this time.

PR #25, opened 2015-10-08
Add tests for product tags

### LINKED ISSUE #24: Add test for product tags
Since we're changing `tags` from a comma-separated string to an `NSSet`, tests need to be added to confirm they work as expected.

runmad:
Fix #24 

### What this does

This adds tests for product tags to confirm that conversion from comma-separated string to an `NSSet` works. Guards against testing products that don't have tags.

Also updated mock tests to include a product with tags.


davidmuzi:
Looks good

PR #38, opened 2015-10-19
Fix shipping rate issue for Apple Pay

### LINKED ISSUE #36: [BUYViewController] failedToUpdateCheckout with error == nil
To reproduce: 
- On Shopify backend, try setting a shipping area to San Francisco.
- On device, choose the default shipping address to be New York
- Experiencing that the Apple Pay process will keep failing to update the checkout
- Only way to solve this is remove New York from the default shipping address and replace with San Francisco instead.

In short, once the shipping address is chosen wrongly, users have to cancel the Apple Pay sheet and start again. Otherwise it will keep failing without specific error given.

### LINKED ISSUE #37: Fix <null> values for shipping rate and shipping rate ID

runmad:
Fix #36
Fix #37 

#### What this does

This fixes an issue with Apple Pay where, if the first address is not a valid shipping address for the shop and no shipping rates are returned, the rates are set to `nil` and it breaks validation on Shopify for all remaining calls to PATCH the `BUYCheckout` object, which would result in a continuous error callback on any subsequent - and valid - address selections.

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Models/BUYCheckout.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Models/BUYCheckout.m
@@ -53,7 +53,7 @@
 
 - (instancetype)initWithCart:(BUYCart *)cart
 {
-	self = [super initWithDictionary:@{}];
+	self = [super init];
 	if (self) {
 		_lineItems = [cart.lineItems copy];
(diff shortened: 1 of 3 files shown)


davidmuzi:
Looks good :ship:

thkeen:
It works. Many thanks

PR #41, opened 2015-10-20
Add NSNull checks and convert to nil

davidmuzi:
Protects from crashing when sending message to NSNull property by converting to nil

@runmad


runmad:
Good to merge.

PR #49, opened 2015-11-03
Apple Pay bug fixes

davidmuzi:
### Fixes

- A discount with a cart that has multiple quantities of the same variant display the incorrect subtotal in the Apple Pay sheet. Note, this just displays the incorrect subtotal amount, the total and charged amount was correct.
- If an Apple Pay sheet is cancelled, the status appears as success, and we did not execute the code to release the reservations on the checkout


runmad:
This looks good. Merge away :+1:

PR #50, opened 2015-11-04
Pop to root after dismissing order Status page

bgulanowski:
When using the "Checkout with Credit Card" option, and reviewing the order in Safari view, the checkout view was not dismissed when returning.

This implements the necessary delegate method to detect when the user is finished and reset to the collections view.

- [ ] test
- [ ] review

@davidmuzi @runmad


bgulanowski:
We need to fix the default branch?

PR #52, opened 2015-11-07
Fix paths so local documentation now links correctly

runmad:
#### What this does:

This reverts the documentation templates as we no longer need custom paths for hosted documentation. Links will now work correctly when the documentation is generated locally and used with apps such as Dash.


davidmuzi:
LGTM

PR #53, opened 2015-11-08
Improved check for default variant

runmad:
### What this does

This adds additional logic to the method that determines whether a product has a single, default variant.


davidmuzi:
Good to merge

PR #54, opened 2015-11-09
Refactor partial address handling to prevent a partial address

davidmuzi:
### Problem

When using Apple Pay, we only receive a partial address which is suitable for retrieving the shipping rates.  Once the purchase is authorized by the user, we receive a full address and need to update the checkout object with Shopify.  Since the `partial_addresses` flag was set the true prior, we need to send the flag back up, but explicitly set to `false` so that Shopify will fail the checkout if the address is not complete (i.e., missing street address).

### Fix

Extract the" partial address" detection into the ApplePayHelper so that its explicit that a partial address will only be sent when an address is updated, and not sent when the purchase is authorized.


runmad:
LGTM.

kiddomk:
Thanks for fixing that. We've been having a lot of troubles with that partial post code returned from Apple Pay. I'm wondering when will you release the next version of the SDK?

davidmuzi:
We have a few other fixes we want to get in, and we will cut a new release

PR #62, opened 2015-11-17
Use bottom layout margins instead of calculating offset using views

runmad:
Improves layout for the checkout footer by using the view controller's bottom layout guide.

@gabrieloc please review :eyes:


davidmuzi:
LGTM

PR #69, opened 2015-12-02
Fix shipping rate invalidation bug using Apple Pay in 1.2.2

### LINKED ISSUE #68: Error: 422 During Checkout - Shipping Rate can't be blank
After upgrading to v1.2.2 of the SDK, we're seeing this error returned when attempting to checkout using Apple Pay.

```
{
    errors =     {
        checkout =         {
            "cart_token" =             (
            );
            email =             (
            );
            "line_items" =             (
            );
            "location_id" =             (
            );
            "reservation_time" =             (
            );
            "shipping_rate_id" =             (
                                {
                    code = blank;
                    message = "can't be blank";
                    options =                     {
                    };
                }
            );
            "source_name" =             (
            );
            "user_id" =             (
            );
        };
    };
}
```

runmad:
### What is this?

Fix #68 

This fixes an issue where the `partialAddresses` flag was set to false, which would invalidate the already set (through a `updateCheckout:completion:` call) shippingRate on Shopify during a call to `completeCheckout:completion:`.

I have added more context to the usage of the `partialAddresses` property and we now override the setter to throw an `NSException` if this is ever set to `false`. This value should always ever be set to `true`.

I have removed the code that would have (and may have) set this property to `false` and added a test assertion that an exception is thrown if the value is set.

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
@@ -101,6 +101,8 @@
 {
 	BUYCart *cart = [[BUYCart alloc] init];
 	BUYCheckout *checkout = [[BUYCheckout alloc] initWithCart:cart];
+	
+	XCTAssertThrows([checkout setPartialAddresses:NO]);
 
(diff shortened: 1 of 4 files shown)


davidmuzi:
Desk reviewed, and tested.

:ship:

PR #73, opened 2015-12-09
Updated API endpoints

runmad:
### What is this?

This PR primarily switches `/anywhere/checkouts` to the new `/api/checkouts` endpoint. However, I thought it was a good chance to switch over to using `NSURLComponents` to create all of the request URLs with, which consisted of some work to use the paths and query items that comes with it.

* `/anywhere/checkouts` -> `/api/checkouts`
* Use of `NSURLComponents` to create request URLs
* methods that take a url have been switched to `NSURL` for better consistency
* updated tests for collections with better assertions for the various components in a URL as well as query comparisons

@bgulanowski please review :eyes: and :tophat:


bgulanowski:
Looks fine. :tophat:

PR #75, opened 2015-12-10
NSURLComponents improvements

runmad:
#### What's this?

* Add additional HTTP method tests
* Improve method names for URL components generation
* Add constants for paths that are used more than once
* `NSDictionary` => `NSQueryItem`s category method can now accept empty strings and correct generate a parameter with a key only
* Using path `pathWithComponents:` and `stringByAppendingPathExtension:` on path creation

@bgulanowski Please review :eyes: and :tophat:


bgulanowski:
:+1: :tophat:

PR #76, opened 2015-12-12
Fix Apple Pay "Setup" button logic

runmad:
### What this does

This moves and adds a few methods in `BUYViewController` from `BUYProductViewController` that handles the logic for whether to display the "Setup  Pay" button if Apple Pay is available on the shop and on the device, but not yet setup with a card.

This also fixes an issue where the "Setup  Pay" would display instead of "Buy with  Pay" in the `BUYProductViewController` due to using an incorrect getter method.


thkeen:
Is it possible to release a version with this pull request? My app just got rejected by Apple due to not displaying the "Setup  Pay". I can fix it manually but would be nice to have it from the SDK.

davidmuzi:
Looks good

runmad:
@thkeen 1.2.4 has been released

PR #79, opened 2015-12-15
Remove unused orderId property

runmad:
### What this does?

This removes an internal property that wasn't being used on the `BUYOrder`. It's always `nil` because it's not parsed or used anywhere.

This removes any confusion and makes it more obvious to use the `BUYObject.identifier`


davidmuzi:
LGTM

PR #85, opened 2016-01-08
Remove shippingRateId on BUYCheckout

### LINKED ISSUE #84: Remove shippingRateId
`shippingRateId` on `BUYCheckout` is deprecated and `BUYShippingRate` takes care of this for us.

Should be removed.

runmad:
Fix #84 

`shipping_rate_id` is no longer required on Shopify. This is a `readonly` property, so there's no need to deprecate it. It can be removed without the need for any fallbacks.

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Models/BUYCheckout.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Models/BUYCheckout.h
@@ -149,11 +149,6 @@
 @property (nonatomic, strong) BUYShippingRate *shippingRate;
 
 /**
- *  Shipping rate identifier
- */
-@property (nonatomic, readonly) NSString *shippingRateId;
-
-/**
  *  A discount added to the checkout
  *  Only one discount can be added to a checkout. Call `updateCheckout:completion:`
  *  after adding a discount to apply the discount code to the checkout.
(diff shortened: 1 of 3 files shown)


Mobile Buy SDK/Mobile Buy SDK/Models/BUYCheckout.h
```diff
@@ -149,11 +149,6 @@
 @property (nonatomic, strong) BUYShippingRate *shippingRate;
 
 /**
- *  Shipping rate identifier
- */
-@property (nonatomic, readonly) NSString *shippingRateId;
```
davidmuzi: In order to maintain backward compatibility, and avoid including the `shippingRateId` in the payload - all we need to remove is the will/didChangeValueForKey in setShippingRateId

davidmuzi: We can mark this property as deprecated, and implement a getter which returns shippingRate.shippingRateIdentifier to maintain backward compatibility

runmad: Added it back and marked as deprecated along with the implementation of a setter.

davidmuzi:
Looks good

PR #86, opened 2016-01-08
Remove province requirement from client-side address validation

### LINKED ISSUE #82: Apple Pay checkout without province field
Hi guys, we’re integrating mobile SDK in many of our Apps (Shopify App: https://apps.shopify.com/mooncode)

But when trying to checkout with an address that does not contain any province (France for instance, but many other countries too), Apple Pay renders the shipping address in red saying that it is not valid.

The method doing this issue is inside "BUYAddress+Additions.h”: “- (BOOL)isValidAddressForShippingRates”.

This method will ALWAYS render NO if no province is entered.

Our workaround (for the moment):

- (BOOL)isValidAddressForShippingRates
{
	BOOL valid = NO;
	
	if (self.city.length > 0 &&
		self.zip.length > 0 &&
//		self.province.length > 0 &&
		(self.country.length > 0 || self.countryCode.length == 2)) {
	
		valid = YES;
	}
	
	return valid;
}

Is there any plan to support countries outside of US / Canada for shipping? We have many customers that need that :) 

Anyway, you’re doing a great job with that SDK!

Amaury

runmad:
Fix #82 

This removed the requirement in Apple Pay for a province to exist on a shipping address. This solves the issue of merchants that ship worldwide and support Apple Pay.

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYAddress+Additions.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYAddress+Additions.m
@@ -46,7 +46,6 @@
 	
 	if (self.city.length > 0 &&
 		self.zip.length > 0 &&
-		self.province.length > 0 &&
 		(self.country.length > 0 || self.countryCode.length == 2)) {
 	
 		valid = YES;


davidmuzi:
LGTM :ship:

Issue #88, opened 2016-01-15
Apple Pay Checkout / Load Shop Race-Condition


cocoahero:
Occasionally, users have reported that upon starting an Apple Pay checkout, they will receive an error message. After looking into it further, it appears it is because of the pre-condition that the shop must be loaded before a checkout can occur. https://github.com/Shopify/mobile-buy-sdk-ios/blob/4c8548c6e8203ae922e29511acd19d53ec125c95/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK/View%20Controllers/BUYViewController.m#L255

In the code for [starting an Apple Pay Checkout](https://github.com/Shopify/mobile-buy-sdk-ios/blob/4c8548c6e8203ae922e29511acd19d53ec125c95/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK/View%20Controllers/BUYViewController.m#L148), it is assumed that because network requests are run serially that the shop information should be loaded prior to a checkout being created. This is some times not the case.

Is there any way to improve this behavior? Perhaps chaining the checkout creation only after a successful shop load?

davidmuzi:
The serial execution of the `getShop` call completing before the `handleCheckout` call is supposed to be enforced by the serial execution of the HTTP requests. https://github.com/Shopify/mobile-buy-sdk-ios/blob/4c8548c6e8203ae922e29511acd19d53ec125c95/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK/Data/BUYClient.m#L94

That said, we have a better, more explicit solution to this using dispatch_groups.  We'll prepare a PR to address this issue, and prevent checkouts from being created if getShop fails.

PR #90, opened 2016-01-20
Prevent race condition during Apple Pay Checkout

davidmuzi:
Fix: #88 

- cleanup error handling for Apple Pay
- Use dispatch_groups to ensure the shop and checkout are completed before continuing with an Apple Pay checkout
- Use same merchant ID in both checkout flows in sample app

@runmad @cocoahero


runmad:
Looks good. I think the check at https://github.com/Shopify/mobile-buy-sdk-ios/blob/bugfix/applepay-race-condition/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK/View%20Controllers/BUYViewController.m#L291 can be removed now since we only call the `requestPayment` method is only called if we have both a shop and a checkout.

PR #103, opened 2016-03-10
Public/apple pay shop fix

### LINKED ISSUE #101: v1.2.5 Breaks Apple Pay Checkout
In an effort to fix a potential race condition where the shop was not loaded before an Apple Pay checkout started, another bug was introduced that from what I can tell, always prevents Apple Pay checkout from succeeding.

https://github.com/Shopify/mobile-buy-sdk-ios/blob/master/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK/View%20Controllers/BUYViewController.m#L150

Shouldn't the above line be `self.shop == nil`??

davidmuzi:
Fix #101 

- Fixes the logic to fetch the BUYShop
- fixes callback thread in Apple Pay helper
- error description change fix
- navbar sizing fix when presenting BUYProductViewController in tabbar
 

Review: @bgulanowski

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Models/BUYError.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Models/BUYError.m
@@ -32,7 +32,7 @@ NSString * const BUYShopifyError = @"BUYShopifyError";
 
 - (NSString *)description
 {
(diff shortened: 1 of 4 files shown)


bgulanowski:
Looks good.

PR #107, opened 2016-03-14
BUYLineItem didn't have a productId when initialized with BUYVariant

poisonedslo:
BUYLineItem didn't have a productId when initialized with BUYVariant


davidmuzi:
This looks good.  Out of curiosity, how is this property being used?  Im wondering if its even necessary since there is a variantId property

poisonedslo:
I'm serializing/deserializing a cart. I didn't find a way to instantiate a
variant with variantId only, is there one?

PR #118, opened 2016-04-07
Update extensions to Cocoa Foundation classes.

bgulanowski:
Standardize category names and unify to have one category per class. Move files to "Additions" directory.

First step in updating to 2.0 model definitions.

cc @davidmuzi @dbart01 @gabrieloc

PR #119, opened 2016-04-08
Reorganize model classes

bgulanowski:
# Purpose

This builds on https://github.com/Shopify/mobile-buy-sdk-ios/pull/118

This primarily relocates some files in preparation for further work.

@dbart01 @davidmuzi @gabrieloc


Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.h
```diff
@@ -62,6 +62,11 @@
 @property (nonatomic, readonly, copy) NSString *productType;
 
 /**
+ * A publically accessible link to the product web page
+ */
+@property (nonatomic, strong) NSURL* publicURL;
```
davidmuzi: This property is no longer exposed, and should be removed

PR #127, opened 2016-04-18
Change customerID type from NSString to the correct type - NSNumber.

### LINKED ISSUE #116: Customer Id on BUYCheckout type is wrong
https://ecommerce.shopify.com/c/shopify-apis-and-technology/t/bug-in-ios-mobile-buy-sdk-cocoapod-1-2-6-checkout-customerid-wrong-class-333589#comment-334208

dbart01:
The `customerId` was being represented by an `NSString` when it is, in fact, returned from the API as an `NSNumber`.

Fixes #116 

@bgulanowski @davidmuzi


bgulanowski:
:shipit:

PR #140, opened 2016-05-04
Improve BUYAccountCredentials API

### LINKED ISSUE #138: Improve BUYAccountCredentials API
The discoverability of available values is somewhat poor. Would be nice to have a more self-documenting interface to make it easy for developers to know exactly what keys are available to use with `BUYAccountCredentialItem`.

dbart01:
#### What this does
- greatly improves discoverability and self-documentation when using `BUYAccountCredentials`
- adds string constants for valid credential item keys
- adds factory methods to eliminate the need to using keys explicitly
- simplifies implementation and validation logic
- 100% test coverage for `BUYAccountCredentials` and `BUYAccountCredentialItem`

Fixes #138 


Mobile Buy SDK/Mobile Buy SDK/Models/BUYAccountCredentials.h
```diff
@@ -25,40 +25,55 @@
...
 
-+ (instancetype)itemWithKey:(NSString *)key value:(NSString *)value;
+@property (nonatomic, assign, readonly, getter=isValid) BOOL valid;
+@property (nonatomic, strong, readonly) NSString *key;
+@property (nonatomic, strong, readonly) NSString *value;
 
-@property (nonatomic, getter=isValid) BOOL valid;
-@property (nonatomic, strong) NSString *key;
-@property (nonatomic, strong) NSString *value;
++ (instancetype)itemEmailWithValue:(NSString *)value;
```
bgulanowski: How about something simpler, like `-itemWithEmail:(NSString *)email`, and similar for the others?

Mobile Buy SDK/Mobile Buy SDK/Models/BUYAccountCredentials.h
```diff
@@ -25,40 +25,55 @@
...
 
-+ (BUYAccountCredentials *)credentialsWithItems:(NSArray<BUYAccountCredentialItem *> *)items;
-+ (BUYAccountCredentials *)credentialsWithItemKeys:(NSArray<NSString *> *)keys;
+@property (nonatomic, assign, readonly) NSUInteger count;
+@property (nonatomic, assign, readonly, getter=isValid) BOOL valid;
+@property (nonatomic, strong, readonly) NSDictionary *JSONRepresentation;
 
-@property (readonly) NSDictionary *JSONRepresentation;
-@property (nonatomic, readonly, getter=isValid) BOOL valid;
++ (BUYAccountCredentials *)credentials;
```
bgulanowski: This is not a useful method.

We should instead add conveniences for the expected use cases: `+credentialsForSignUp`, `+credentialsForLogIn`, `+credentialsForReset`, and `-credentialsForActivation`.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYAccountCredentials.m
```diff
@@ -26,94 +26,133 @@
...
-		BUYAccountCredentialItem *item = [BUYAccountCredentialItem itemWithKey:key value:@""];
-		[items addObject:item];
+	NSAssert(items, @"BUYAccountCredentialItem array cannot be nil.");
+	for (BUYAccountCredentialItem *item in items) {
+		[self setCredentialItem:item];
 	}
-	return [BUYAccountCredentials credentialsWithItems:items];
 }
 
+- (void)setCredentialItem:(BUYAccountCredentialItem *)item
```
bgulanowski: I'm thinking that we should make this class immutable after initialization.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYAccountCredentials.h
dbart01: I like that, will change.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYAccountCredentials.m
dbart01: I completely agree with immutability here. The only reason I kept it is to maintain a similar API to what it was before. Otherwise, I'm happy to make this an immutable class.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYAccountCredentials.h
dbart01: How are values assigned in these cases if both the `BUYAccountCredentials` and `BUYAccountCredentialItem` are immutable? If go the immutable route, I imagine we'll only have on initialize (plus convenience class initializer), no?

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
```diff
@@ -327,14 +327,17 @@ - (void)testProductsInCollectionWithSortOrderTitleDescending
...
-	NSURLSessionDataTask *task = [self.client createCustomerWithCredentials:credentials callback:nil];
+	BUYAccountCredentials *credentials = [BUYAccountCredentials credentialsWithItems:@[
+																					   [BUYAccountCredentialItem itemWithFirstName:@"michael"],
+																					   [BUYAccountCredentialItem itemWithLastName:@"scott"],
+																					   [BUYAccountCredentialItem itemWithEmail:@"fake@example.com"],
+																					   [BUYAccountCredentialItem itemWithPassword:@"password"],
+																					   [BUYAccountCredentialItem itemWithPasswordConfirmation:@"password"],
+																					   ]];
+	
+	NSURLSessionDataTask *task = [self.client createCustomerWithCredentials:credentials callback:^(BUYCustomer * _Nullable customer, NSString * _Nullable token, NSError * _Nullable error) {
```
bgulanowski: Why do we need these empty blocks?

Also, I find the nullability modifiers really clutter up the implementation. They may be useful as reminders, but we don't really need them.

dbart01: Not sure why it shows up with this excessive indentation. It looks more civilized in Xcode. The nullability annotation is inserted by code completion so it wasn't intentional, but I agree its way more cluttered with it. Will fix both.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
```diff
@@ -327,14 +327,18 @@ - (void)testProductsInCollectionWithSortOrderTitleDescending
...
+					   [BUYAccountCredentialItem itemWithFirstName:@"michael"],
+					   [BUYAccountCredentialItem itemWithLastName:@"scott"],
+					   [BUYAccountCredentialItem itemWithEmail:@"fake@example.com"],
+					   [BUYAccountCredentialItem itemWithPassword:@"password"],
+					   [BUYAccountCredentialItem itemWithPasswordConfirmation:@"password"],
+					   ];
+	BUYAccountCredentials *credentials = [BUYAccountCredentials credentialsWithItems:items];
+	
+	NSURLSessionDataTask *task = [self.client createCustomerWithCredentials:credentials callback:^(BUYCustomer *customer, NSString *token, NSError *error) {
+
```
bgulanowski: These empty blocks are bugging me.

We should add an expectation and fulfill it when the block is called. We should also assert that we get the right arguments.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.h
```diff
@@ -163,3 +164,5 @@ typedef void (^BUYDataOrdersBlock)(NSArray <BUYOrder*> *orders, NSError *error);
 - (NSURLSessionDataTask *)getOrdersForCustomerWithCallback:(BUYDataOrdersBlock)block;
 
 @end
+
+NS_ASSUME_NONNULL_END
```
bgulanowski: Missing newline.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.h
```diff
@@ -36,7 +37,7 @@
  *  @param customer A BUYCustomer
  *  @param error    An optional NSError
  */
-typedef void (^BUYDataCustomerBlock)(BUYCustomer *customer, NSError *error);
+typedef void (^BUYDataCustomerBlock)(BUYCustomer * _Nullable customer, NSError * _Nullable error);
```
bgulanowski: We could make these declarations easier to parse by using assume nullable macros around them, and move the assume nonnull begin macro further down. Just a suggestion, though. I'm 50/50 on which is better.

bgulanowski:
A couple more suggestions/observations. Use your discretion. Merge when ready.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.h
dbart01: As far as I know, there is no `NS_ASSUME_NULLABLE` or similar macro . Just `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
dbart01: These blocks are empty only to satisfy the `nonnull` constraint for the callback parameter. Prior to the introduction of that constraint we just passed `nil`.

We can't setup an expectation here because the callbacks aren't functional. They never get called because we're using a subclass of `BUYClient` that never resumes the network call task. I'm open to other suggestions for getting rid of the empty blocks but for now, this is the best I could come up with.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.h
bgulanowski: Unfortunate.

PR #150, opened 2016-05-09
Fix broken file after merge.

bgulanowski:
This reverts a file which was accidentally committed with conflicting changes after a merge, even though it was only supposed to be changed locally.

@davidmuzi


davidmuzi:
:shipit:

Issue #143, opened 2016-05-06
Write migration guide


bgulanowski:
Users switching to 2.0 from 1.x need a focussed guide for specific API changes.

These changes are made in another repo, private to Shopify.

zebarahman:
Hi, I need to add the Customer API into my app. How do I do it? Do I have to wait for version 2.0 release? When will that be? Thanks.

PR #151, opened 2016-05-09
Add Customer Sample App README.md

dbart01:
### What this is

We'll need a readme file for the customer sample app. This adds one.

@bgulanowski @davidmuzi


davidmuzi:
Are we going to have a more descriptive integration guide for using the Customer API?  There are nuances that should be documented such as storing the customer token, handing token refreshing, and handling password reset and account activation in-app, for instance.

bgulanowski:
There is an issue here: https://github.com/Shopify/mobile-buy-sdk-ios/issues/143

dbart01:
As @bgulanowski mentioned, there will be a separate guide. Since these nuances aren't showcased in the sample app, its best that the readme reflects that, in my opinion.

bgulanowski:
👍

PR #152, opened 2016-05-09
Update readme.

### LINKED ISSUE #142: Update readme.md with Customer API docs
The README should document some key features of the new Customer API exposed in the SDK.

dbart01:
### What this does
After adding customer API support to SDK 2.0, the home page readme file has become out of date. In this PR, I update the storefront API example to include fetching collections and products as well as add new example for customer endpoints that show login and customer creation.

It's best reviewed formatted here: https://github.com/Shopify/mobile-buy-sdk-ios/blob/feature/update-readme/README.md

Fixes #142 

@bgulanowski @davidmuzi

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,11 @@ We can now use the credentials object to login the customer.
 Creating a customer account is very similar to login flow but requres a little more info in the credentials object.
 ```objc
 NSArray *credentialItems = @[
-								 [BUYAccountCredentialItem itemWithFirstName:@"John"],
-								 [BUYAccountCredentialItem itemWithLastName:@"Smith"],
-								 [BUYAccountCredentialItem itemWithEmail:@"john.smith@gmail.com"],
-								 [BUYAccountCredentialItem itemWithPassword:@"password"],
-								 [BUYAccountCredentialItem itemWithPasswordConfirmation:@"password"],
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,20 +96,74 @@ Initialize the `BUYClient` with your credentials from the *Mobile App Channel*
...
+	if (products && !error) {
+		// Do something with products
+	} else {
+		NSLog(@"Error fetching products in collection %@: %@", collection.collectionId, error.userInfo);
+	}
 }];
 ```
-
 Consult the [Usage Section](https://docs.shopify.com/mobile-buy-sdk/ios/integration-guide/#using-the-mobile-buy-sdk) of the Integration Guide on how to create a cart, and checkout with the SDK.
+### Customer API
```
bgulanowski: The line above which links to the guide for cart/checkout should have a section header, otherwise it seems to be part of the section above.

bgulanowski: Oh, wait, I suppose "Storefront" is meant to include both catalog and checkout. That's probably OK, although it might be nice to separate them.

dbart01: Yeah, it was part of that section initially but you're right, it should probably be a separate section.

bgulanowski:
:shipit:

PR #153, opened 2016-05-10
Add payment providers

bgulanowski:
# Why?

We want to consolidate payment logic and ensure consistency between different payment types.

# How?

Introduce a payment controller which manages many payment providers, each of which designed to interact with a payment interface that provides the necessary payment workflow or data to complete a checkout.

@dbart01 @davidmuzi


Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYApplePayPaymentProvider.h
```diff
@@ -0,0 +1,75 @@
...
+ *  The supported credit card payment networks.  Default values:
+ *  iOS 8.3: PKPaymentNetworkAmex, PKPaymentNetworkMasterCard, PKPaymentNetworkVisa.
+ *  iOS 9.0: PKPaymentNetworkAmex, PKPaymentNetworkDiscover, PKPaymentNetworkMasterCard, PKPaymentNetworkVisa.
+ */
+@property (nonatomic, copy) NSArray * _Nonnull supportedNetworks;
+
+/**
+ *  The merchant ID required for Apple Pay
+ */
+@property (nonatomic, copy, readonly) NSString * _Nonnull merchantID;
```
dbart01: Perhaps we could wrap the header in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` ? It seems most values are non nil.

Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYApplePayPaymentProvider.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ *  Initializer for Apple Pay payment provider
+ *
+ *  @param client     a `BUYClient`
+ *  @param merchantID the merchant ID for Apple Pay
+ *
+ *  @return an instance of `BUYApplePayPaymentProvider`
+ */
+- (instancetype _Nullable)initWithClient:(BUYClient * _Nonnull)client merchantID:(NSString * _Nonnull)merchantID NS_DESIGNATED_INITIALIZER;
```
dbart01: This initializer never returns nil but is marked `_Nullable`.

Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYPaymentProvider.h
```diff
@@ -0,0 +1,141 @@
...
+ */
+- (void)paymentProvider:(id <BUYPaymentProvider>)provider didFailToUpdateCheckoutWithError:(NSError *)error;
+
+/**
+ *  Called when the checkout failed
+ *
+ *  @param provider   the `BUYPaymentProvider`
+ *  @param error    the optional `NSError`
+ */
+- (void)paymentProvider:(id <BUYPaymentProvider>)provider didFailCheckoutWithError:(NSError *)error;
```
davidmuzi: Can we collapse the error methods since the NEError can convey the actual error condition?

Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYPaymentProvider.h
```diff
@@ -0,0 +1,141 @@
...
+ *  @param provider   the `BUYPaymentProvider`
+ */
+- (void)paymentProviderWillStartCheckout:(id <BUYPaymentProvider>)provider;
+
+/**
+ *  Called when the checkout has been dismissed
+ *
+ *  @param provider   the `BUYPaymentProvider`
+ */
+- (void)paymentProviderDidDismissCheckout:(id <BUYPaymentProvider>)provider;
```
davidmuzi: Can we get rid of this too?

Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYWebCheckoutPaymentProvider.m
```diff
@@ -0,0 +1,193 @@
...
+
+- (NSString *)identifier
+{
+	return BUYWebPaymentProviderId;
+}
+
+- (void)startCheckout:(BUYCheckout *)checkout
+{	
+	if (self.isInProgress) {
+		return;
```
davidmuzi: Should we call the error delegate method, and/or log an error here?  Im thinking the delegate may not be the right use here since it may be interpreted that the original checkout failed.

Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYPaymentProvider.h
bgulanowski: In our own app, we implement both, but effectively ignore the "update failed" error. They seem different enough to me. The alternative is that the delegate has to implement switching logic in the error handler. Switching logic makes code harder to read.

Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYWebCheckoutPaymentProvider.m
bgulanowski: We should throw an exception, actually.

Mobile Buy SDK/Mobile Buy SDK Tests/test_shop_data.json
```diff
@@ -5,8 +5,8 @@
 	"channel_id": "",
 	"app_id": "",
 	"merchant_id": "",
-	"customer_email": "",
-	"customer_password": "",
+	"customer_email": "asd@asd.com",
+	"customer_password": "asdasd",
```
gabrieloc: This probably shouldn't have been committed.

PR #160, opened 2016-05-11
Improve checkout API

### LINKED ISSUE #156: Add convenience checkout API
Provide convenience methods on `BUYClient` to implement the full standard checkout complete flow, once a payment has been completed.

- complete the checkout
- poll on the status
- get the checkout with the order
- return checkout or error

It would be ideal if we had a single method that accepted an object that wrapped an Apple Pay token or a card server token. That object should generate the correct payload on behalf of the `BUYClient`, to avoid duplicate code in the `BUYClient`.

cc @krisorr, @davidmuzi, @dbart01

### LINKED ISSUE #162: Refactor client API
Clean up `BUYClient` API and make it Swift-friendly with nullability annotations.

dbart01:
### What this does

- adds two classes conforming to BUYPaymentSessionProvider for internal use
- modifies the API to only return payment tokens wrapped by protocol
- merges two `completePayment` methods into a single method
- refactors methods to use `NSAssert` instead of returning `nil` or an error object
- adds nullability annotation to `BUYClient`
- adds tests for  `id<BUYPaymentToken>` concrete classes

Fixes #162 

@bgulanowski @davidmuzi


Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
```diff
@@ -444,37 +446,20 @@ typedef void (^BUYDataGiftCardBlock)(BUYGiftCard *giftCard, NSError *error);
...
- *  You must get the job's status by calling checkCompletionStatusOfCheckout:block
- *
- *  Note: There's no guarantee that the BUYCheckout returned will be the same as the one that is passed in.
- *  We recommended using the BUYCheckout returned in the block.
- *
- *  @param checkout The BUYCheckout to complete
- *  @param token    The PKPaymentToken
- *  @param block    (^BUYDataCheckoutBlock)(BUYCheckout *checkout, NSError *error);
+ *  @param checkout      The BUYCheckout to complete
+ *  @param paymentToken  Opaque payment token object
```
bgulanowski: There is no such param as `paymentToken` on the `BUYDataCheckoutBlock` block type signature.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -424,75 +420,35 @@ - (NSURLSessionDataTask *)getCheckout:(BUYCheckout *)checkout completion:(BUYDat
 
 - (NSURLSessionDataTask *)updateCheckout:(BUYCheckout *)checkout completion:(BUYDataCheckoutBlock)block
 {
+	NSAssert(checkout, @"Failed to update checkout. Checkout must not be nil");
+	NSAssert([checkout hasToken], @"Failed to update checkout. Checkout must have a valid token associated with it/");
```
bgulanowski: Typo at the end of the assert comment. I think we can get away with just the second assert, as the first is implicit in it.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -574,34 +526,30 @@ - (NSURLSessionDataTask *)getShippingRatesForCheckout:(BUYCheckout *)checkout co
 
 #pragma mark - Payments
 
-- (NSURLSessionDataTask *)storeCreditCard:(BUYCreditCard *)creditCard checkout:(BUYCheckout *)checkout completion:(BUYDataCreditCardBlock)block
+- (NSURLSessionDataTask *)storeCreditCard:(BUYCreditCard *)creditCard checkout:(BUYCheckout *)checkout completion:(BUYDataCreditCardBlock)completion
 {
-	NSURLSessionDataTask *task = nil;
+	NSAssert(checkout, @"Failed to store credit card. No checkout provided.");
+	NSAssert(checkout.token, @"Failed to store credit card. No checkout token provided.");
```
bgulanowski: Maybe we need a macro that makes this assert simpler to include. This one is inconsistent with others.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYCreditCardToken.m
```diff
@@ -0,0 +1,51 @@
...
+	self = [super init];
+	if (self) {
+		_paymentSessionID = paymentSessionID;
+	}
+	return self;
+}
+
+#pragma mark - BUYPaymentSessionProvider -
+
+- (NSDictionary *)jsonRepresentation
```
bgulanowski: I think "JSON" should be all caps, and I wonder if we should use "JSONDictionary" as found on `BUYObject`.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
dbart01: That parameters isn't for `BUYDataCheckoutBlock`. Its for the `completeCheckout:paymentToken:completion:` method below.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYCreditCardToken.m
dbart01: Renamed.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
dbart01: Converted to single assert.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
bgulanowski: Gah, the diff plus I don't read good. Sorry :)

dbart01:
PR is ready for review (sorry about the size).

Mobile Buy SDK/Mobile Buy SDK Tests/BUYApplePayTokenTests.m
```diff
@@ -0,0 +1,71 @@
+//
+//  BUYApplePayTokenTests.m
+//  Mobile Buy SDK
+//
+//  Created by Dima Bart on 2016-05-12.
+//  Copyright © 2016 Shopify Inc. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+#import <PassKit/PassKit.h>
```
davidmuzi: Its preferred to use @imports

Mobile Buy SDK/Mobile Buy SDK Tests/BUYApplePayTokenTests.m
```diff
@@ -0,0 +1,71 @@
...
+
+@property (strong, nonatomic) NSData *testData;
+
+@end
+
+@implementation BUYApplePayTestToken
+
++ (instancetype)validToken {
+	BUYApplePayTestToken *token = [BUYApplePayTestToken new];
+	token->_testData            = [BUYTestingToken dataUsingEncoding:NSUTF8StringEncoding];
```
davidmuzi: why is the property accessor not used here?

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
```diff
@@ -74,7 +75,7 @@ - (NSData *)dataForCartFromClient:(BUYClient *)client
 {
 	BUYCart *cart = [self cart];
 	BUYCheckout *checkout = [[BUYCheckout alloc] initWithCart:cart];
-	NSURLSessionDataTask *task = [self.client createCheckout:checkout completion:nil];
+	NSURLSessionDataTask *task = [self.client createCheckout:checkout completion:^(BUYCheckout *checkout, NSError *error) {}];
```
davidmuzi: why are all these empty blocks required?

Mobile Buy SDK/Mobile Buy SDK Tests/BUYApplePayTokenTests.m
dbart01: Perhaps, but we don't use them for any other modules like `Foundation` or `UIKit` or `XCTest`. I guess I just prefer the consistency.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYApplePayTokenTests.m
dbart01: Left over from previous experiment. Will fix.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest.m
dbart01: Because of the `nonnull` specifier. Otherwise it emits a warning, which is treated as an error.

davidmuzi: Thats fair, since most of the APIs return vital payload - callers should expect to have a callback

dbart01: Precisely, since the callback is the only way for the functions to return data, they make no sense if a callback isn't provided.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYIntegrationTest.m
```diff
@@ -988,8 +992,9 @@ - (void)testCheckoutFlowUsingCreditCard
 	[self createCheckout];
 	[self fetchShippingRates];
 	[self updateCheckout];
-	[self addCreditCardToCheckout];
-	[self completeCheckout];
+	
+	id<BUYPaymentToken> token = [self addCreditCardToCheckout];
```
bgulanowski: Nitpick: local var is probably not needed.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
```diff
@@ -444,37 +445,20 @@ typedef void (^BUYDataGiftCardBlock)(BUYGiftCard *giftCard, NSError *error);
 - (NSURLSessionDataTask *)updateCheckout:(BUYCheckout *)checkout completion:(BUYDataCheckoutBlock)block;
 
 /**
- *  Finalizes the BUYCheckout and charges the credit card.
+ *  Finalizes the BUYCheckout and charges the payment provider (ex: Credi Card, Apple Pay, etc).
```
bgulanowski: Typo: "Credi Card" (missing "t")

Mobile Buy SDK/Mobile Buy SDK/Models/BUYPaymentSessionProvider.h
```diff
@@ -0,0 +1,31 @@
+//
+//  BUYPaymentToken.h
```
bgulanowski: This file still has the old name.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -57,6 +55,8 @@
 #define kMinSuccessfulStatusCode 200
 #define kMaxSuccessfulStatusCode 299
 
+#define BUYAssertCheckout(checkout) NSAssert([(checkout) hasToken], @"Checkout assertion failed. Checkout must have a valid token associated with it.")
```
bgulanowski: We should throw an explicit exception, because `NSAssert()` is disabled in release builds, but we should always complain/crash if we are passed an invalid checkout.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -141,8 +143,10 @@ - (NSURLSessionDataTask *)getShop:(BUYDataShopBlock)block
 - (NSURLSessionDataTask *)getProductsPage:(NSUInteger)page completion:(BUYDataProductListBlock)block
 {
 	NSURLComponents *components = [self URLComponentsForChannelsAppendingPath:kBUYClientPathProductPublications
-																   queryItems:@{@"limit" : [NSString stringWithFormat:@"%lu", (unsigned long)self.pageSize],
-																				@"page" : [NSString stringWithFormat:@"%lu", (unsigned long)page]}];
+																   queryItems:@{
+																				@"limit" : @(self.pageSize).stringValue,
```
bgulanowski: 👍  nice

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -334,65 +345,48 @@ - (NSURLSessionDataTask *)createCheckoutWithCartToken:(NSString *)cartToken comp
 
 - (NSURLSessionDataTask *)postCheckout:(NSDictionary *)checkoutJSON completion:(BUYDataCheckoutBlock)block
 {
-	NSURLSessionDataTask *task = nil;
-	NSError *error = nil;
-	NSData *data = [NSJSONSerialization dataWithJSONObject:checkoutJSON options:0 error:&error];
+	NSData *data = [NSJSONSerialization dataWithJSONObject:checkoutJSON options:0 error:nil];
```
bgulanowski: This method is not public API, so I don't believe it should use asserts.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -424,75 +419,33 @@ - (NSURLSessionDataTask *)getCheckout:(BUYCheckout *)checkout completion:(BUYDat
...
-		NSError *error = nil;
-		NSData *data = [NSJSONSerialization dataWithJSONObject:paymentJson options:0 error:&error];
-		if (data && !error) {
-			task = [self checkoutCompletionRequestWithCheckout:checkout body:data completion:block];
-		}
-		else {
-			block(nil, [NSError errorWithDomain:kShopifyError code:BUYShopifyError_InvalidCheckoutObject userInfo:nil]);
-		}
-	}
+	NSAssert(data || isFree, @"Failed to complete checkout. Checkout must have a payment token or have a payment value equal to $0.00");
```
bgulanowski: I prefer that all asserts on arguments are at the very top of the method, for easier reading. In this case, `data` depends on `paymentToken`, so we should just assert (or throw invalid arguments exception) that `paymentToken` is valid. Could add a read-only `valid (getter=isValid)` property to `BUYPaymentToken`.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -553,67 +497,46 @@ - (NSURLSessionDataTask *)getCompletionStatusOfCheckoutToken:(NSString *)token c
...
-		}
-		else {
-			block(nil, nil, [NSError errorWithDomain:kShopifyError code:BUYShopifyError_InvalidCheckoutObject userInfo:nil]);
-		}
-	}
-	return task;
+	
+	NSData *data = [NSJSONSerialization dataWithJSONObject:@{ @"checkout" : json } options:0 error:nil];
+	
+	NSAssert(data, @"Failed to store credit card. Unable to serialize JSON payload. Possibly invalid credit card object.");
```
bgulanowski: This assert seems superfluous, since this method controls the creation of the `json` and `data`.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -646,14 +569,12 @@ - (NSError *)errorFromJSON:(NSDictionary *)errorDictionary statusCode:(NSInteger
...
-	NSData *data = [NSJSONSerialization dataWithJSONObject:json options:0 error:&error];
-	NSURLSessionDataTask *task = nil;
-	if (data && !error) {
-		task = [self requestForURL:url method:method body:data completionHandler:completionHandler];
-	}
-	return task;
+	NSAssert(object, @"Failed to perform request. id<BUYSerializable> must not be nil.");
+	
+	NSData *data = [NSJSONSerialization dataWithJSONObject:[object jsonDictionaryForCheckout] options:0 error:nil];
+	NSAssert(data, @"Failed to perform request. Could not serialize object. Possibly invalid object.");
```
bgulanowski: As above, assert the inputs.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYApplePayToken.m
```diff
@@ -0,0 +1,63 @@
...
+
+@implementation BUYApplePayToken
+
+#pragma mark - Init -
+
+- (instancetype)initWithPaymentToken:(PKPaymentToken *)paymentToken
+{
+	self = [super init];
+	if (self) {
+		NSAssert(paymentToken.paymentData.length > 0, @"Failed to initialize BUYApplePayToken. Invalid or nil paymentToken.");
```
bgulanowski: Move the assert up to the first line of the method.

bgulanowski:
Another thing that could be improved in the `BUYClient` is the inconsistent strategy for converting input models into JSON and `NSData`. There seems to be unnecessary duplication in this regard.

For example, `-checkoutCompletionRequestWithCheckout:body:` could be revised to `-checkoutCompletionRequestWithCheckout:object:` where object is `id<BUYSerializable>`.

Similarly, we could probably remove `-patchRequestForURL:body:completionHandler:` (and `-putRequestForURL:…`) since we already have `-patchRequestForURL:object:completionHandler:`.

bgulanowski:
It might also be good if `id<BUYSerializable>` objects could return `NSData` directly, saving the `BUYClient` from the JSON->data transformation.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
```diff
@@ -644,23 +555,56 @@ - (NSError *)errorFromJSON:(NSDictionary *)errorDictionary statusCode:(NSInteger
...
+}
+
+- (NSURLSessionDataTask *)requestForURL:(NSURL *)url method:(NSString *)method object:(id <BUYSerializable>)object completionHandler:(void (^)(NSDictionary *json, NSURLResponse *response, NSError *error))completionHandler
+{
+	NSData *data = nil;
+	if (object) {
+		data = [NSJSONSerialization dataWithJSONObject:[object jsonDictionaryForCheckout] options:0 error:nil];
+	}
+	
+	return [self requestForURL:url method:method body:data completionHandler:completionHandler];
```
bgulanowski: This is now the only method that calls the "body/data" version. Should we merge them?

bgulanowski:
This looks really good to me. There are more opportunities to refactor, but they are all internal and can be done separately, at a later date. This gets the new exception-throwing behaviour in place, standardizes the interfaces, and gets us ready for the revised checkout API.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
dbart01: Good idea, will do.

dbart01:
Good to merge, then?

bgulanowski:
Was waiting to see if there were any more comments, but as far as I'm concerned, 🎩

PR #165, opened 2016-05-13
Refactor BUYClient to remove creation and management of routes

### LINKED ISSUE #164: Move API route creation and management out of BUYClient
The `BUYClient` should know nothing about creating and managing routes used for API access. The work should be delegated to a separate object, tested separately.

dbart01:
### Why

Previously, the `BUYClient` was responsible for knowing and creating API routes. As a result, it bloated the class. Following the single-responsibility principle, this work has been moved to a separate class - `BUYRouter`. The router vends `BUYRoute` object, which have optional customizable query items for configuring the URL.

### What
- separates `BUYClient` components into separate category file
- moves URL creation (routing) into a `BUYClient` category
- adds dedicated test coverage for URL creation (routing)
- remove unnecessary import, static string declarations, messy macro and other miscellaneous cleanup

Fixes #164 
Relies on #160

@bgulanowski @davidmuzi


dbart01:
This is a work in progress. I'd like to get some feedback regarding this approach in general before I start writing tests for `BUYRouter` / `BUYRoute`.

davidmuzi:
The router class appears to re-implement a lot of the functionality we got from NSURLComponents.  Im in favour for breaking up responsibility, but couldn't we have moved the NSURLComponent factory methods instead of re-implementing them?  
I was thinking that the BUYClient class could instead be broken up into having each entity represented by a category, i.e., BUYClient+Customers, BUYClient+Checkout, etc.

dbart01:
The reason for the reimplementation is that the `NSURLComponents` were very hard to follow, we couldn't see all API endpoint we were hitting all in one class and it wasn't testable. The `BUYRouter` and `BUYRoute` uses `NSURLComponents` under the hood and abstracts away that logic. If we decide to switch away from `NSURLComponents` in the future or if Apple makes changes to these classes or introduces a new one (they do that ofter), it won't break the rest of the API. Its a much more robust approach, in my opinion.

bgulanowski:
`BUYRouter` isn't literally routing anything. Is there another name that succinctly captures the role of URL generation?

I think the `BUYRoute` class is not adding anything. The `BUYClient` shouldn't even care about `NSURLComponents`. It only cares about URLs.

It would be really cool if the various models were responsible for generating the URLs, still using URL components internally. These could be written as category methods in one place. While the models shouldn't know about the other parts of the URL (the scheme, the domain, the root path), there is no reason for them to not know how to generate a relative path. Another object (the URL generator) can hold that info, and merge it with the path from the model object.

Just an idea. I'm open to alternatives.

davidmuzi:
Recap of IRL discussion:
- Break up methods for each type of entity in a category of BUYClient. (Ie, BUYClient+Customers, BUYClient+Checkout, etc)
- Convert the URL Construction here into a category on BUYClient (BUYClient+Routes) which returns NSURLs

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
```diff
@@ -0,0 +1,57 @@
...
+//
+
+#import "BUYClient.h"
+
+@interface BUYClient (Routing)
+
+- (NSURL *)routeForAPI;
+- (NSURL *)routeForApps;
+
+- (NSURL *)routeForShop;
```
davidmuzi: can these be renamed to <#noun#>URL.  

ex. shopURL, CustomersOrdersURL, etc...

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.m
```diff
@@ -0,0 +1,222 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+#import "BUYClient+Routing.h"
+
+#pragma mark - NSURL (Private Routing) -
+
+@interface NSURL (PrivateRouting)
```
davidmuzi: doesn't look like this is needed

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
```diff
@@ -114,104 +114,12 @@ typedef NS_ENUM(NSUInteger, BUYStatus) {
 };
 
 /**
```
davidmuzi: This header should not need BUYCollectionSort, BUYClientCustomerAccessToken, BUYDataStatusBlock, BUYStatus, and several of the @class forward declarations

davidmuzi:
Great clean-up.  Some comments, and make sure the appropriate headers are public so the sample apps and unit tests run.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
dbart01: I find having the `routeFor` prefix provides very nice code-completion. It basically acts as a name space. Are we sure we don't want to take advantage of that?

davidmuzi: Valid point, but these are only consumed in the client implementation and
are not intended for public consumption

dbart01: I think code-completion benefit is great for internal use as well. Perhaps we can rename the methods to something like `- (NSURL *)URLFor...`?

Mobile Buy SDK/Mobile Buy SDK Tests/test_shop_data.json
```diff
@@ -5,8 +5,8 @@
 	"channel_id": "",
 	"app_id": "",
 	"merchant_id": "",
-	"customer_email": "",
-	"customer_password": "",
+	"customer_email": "asd@asd.com",
+	"customer_password": "asdasd",
```
dbart01: This is necessary for local test to run properly. Perhaps in the future, we can investigate rewriting tests to run without this requirement.

davidmuzi: This file is shared with Android - ensure they don't get out of sync

dbart01: I think we had these values initially but they got lost in rebase / merge conflicts (I think). How is this looking for Android @krisorr ?

krisorr:
That matches the email and password we have in the code.  It is not currently in our test_shop_data.json, I believe that is TBD

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
davidmuzi: thats fine

dbart01:
All comments have been addressed. Ready for re-review.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
```diff
@@ -78,10 +66,6 @@ typedef NS_ENUM(NSUInteger, BUYCollectionSort) {
 	BUYCollectionSortCreatedDescending
```
davidmuzi: BUYCollectionSort shouldn't be in this header

davidmuzi:
🚢

PR #167, opened 2016-05-16
Migrate to NSOperation-based network requests.

### LINKED ISSUE #166: Introduce NSOperation-based networking
Currently, every API call return an `NSURLSessionTask` handle for the network operation. For consistency across the API and to provide operation dependency management, we'll need a set of `NSOperation` subclasses to handle networking (using `NSURLSession` under the hood).

Doing so will also provide support for cancellation in aggregate API calls that call multiple endpoints (ex: Convenience checkout API #156)

dbart01:
### What this does
- adds ` BUYRequestOperation` and `BUYOperation` classes
- wraps all `NSURLSessionTask`-based requests in a `BUYRequestOperation`
- 100% test coverage for `BUYRequestOperation` and underlying `BUYOperation`
- adds an explicit `requestQueue` property to `BUYClient` where all request operations and their underlying `NSURLSessionDataTask`s are executed

Fixes #166 
Depends on #165 

@bgulanowski @davidmuzi @gabrieloc @dengjeffrey


bgulanowski:
First look, this looks really great. Some concerns about whether the properties of the custom operations are fully atomic, because I think they should be. Look into `OSAtomic` functions for getting/setting primitive values atomically.

Was going to point to reference docs but this is probably better: https://mikeash.com/pyblog/friday-qa-2011-03-04-a-tour-of-osatomic.html -- I haven't read it, but Mike Ash's reputation is platinum.

davidmuzi:
Please make sure all sample apps continue to build before opening a PR that introduces breaking changes

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYOperation.m
```diff
@@ -0,0 +1,132 @@
...
+{
+	self = [super init];
+	if (self) {
+		_lock = [NSLock new];
+	}
+	return self;
+}
+
+#pragma mark - Concurrent -
+- (BOOL)isAsynchronous
```
davidmuzi: these 2 methods are not required when using queues.

"When you add an operation to an operation queue, the queue ignores the value of the asynchronous property and always calls the start method from a separate thread. Therefore, if you always run operations by adding them to an operation queue, there is no reason to make them asynchronous."

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
```diff
@@ -240,6 +241,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nonnull) BUYModelManager *modelManager;
 
 /**
+ *  Queue where on which all request operation are executed
+ */
+@property (nonatomic, strong, readonly, nonnull) NSOperationQueue *requestQueue;
```
davidmuzi: for consistency, lets type the callback queue to an NSOperation, and name it `callbackQueue` for clarity.

then when you need to callback, you can:
```
		[callbackQueue addOperationWithBlock: ^{
			completionHandler(json, response, error);
		}];
```

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYOperation.m
dbart01: Agreed. While not strictly necessary for use with operation queues, I think we should keep this method around for conformance the spec outlined in the doc:

> If you are creating a concurrent operation, you need to override the following methods and properties at a minimum:
> - start
> - asynchronous
> - executing
> - finished

Granted, we can remove `- (void)isConcurrent` since is deprecated.

dbart01:
@bgulanowski Regarding atomic properties, is there anything specific you had in mind? From my understanding, the synchronized access to the `state` property via an `NSLock` is enough to ensure thread-safety on for `BUYOperation` KVO-compliant state changes. I could be wrong, though.

bgulanowski:
The `state` property should be declared `atomic`. But instead of writing an atomic getter, you should remove your override of the setter.

Instead, declare KVO dependencies using `+keyPathsAffectingIsExecuting` and `+keyPathsAffectingIsFinished`, and have those both return `[NSSet setWithObject:@"state"]`. And then implement getters for `-isFinished` and `-isExecuting`.

Then KVO will/did change notifications will automatically be sent for both `isExecuting` and `isFinished` whenever `state` is changed. And since technically, both of them do change when `state` changes, that's the right behaviour (not just one or the other).

dbart01:
I have some concerns with this approach:
- the threading behaviour isn't explicit, which can (and it often does) be overlooked when the code changes some time in the future
- we're relying on built-in `KVO` implementation which is somewhat obscure and the code path isn't clear just by looking at the class file

The main concern, however, is with `KVO` notifications and thread-safety. Can two thread simultaneously change the `state` property and introduce a condition where the `KVO` notifications are triggered twice? I think explicitly synchronizing the `KVO` notifications is a safer bet.

Also, since `isExecuting` and `isFinished` are mutually exclusive states, it isn't necessary to post `KVO` notifications for both, although it would happen anyway as the setter posts notifications for `oldPath` and `newPath` while holding a lock.

BTW, the idea was borrowed from the `AFNetworking` implementation of concurrent `NSOperation` subclasses back from v2.0. Check it out: https://github.com/AFNetworking/AFNetworking/blob/2.x/AFNetworking/AFURLConnectionOperation.m#L322-L337

bgulanowski:
Notifications sent manually are no different than those sent automatically by properly atomic methods. But atomic properties should be declared and fully implemented as such (setter as well as getter). And derived properties should have their getters implemented explicitly.

The implementation of the getters for the derived properties should not need to use locks. The underlying property accessors, being atomic, will prevent changes to the property (and thus the derived properties) within observers' implementations of `-observeValueForKeyPath:…`.

I also think that we could drop the enum altogether in favour of a simple `BOOL` property, but I may be wrong. We can discuss more later today.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYOperation.h
```diff
@@ -0,0 +1,37 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface BUYOperation : NSOperation
```
bgulanowski: This class needs dedicated tests.

bgulanowski:
If we are confident that this implementation is reliable, which requires tests, we can consider alternative implementations in future.

dbart01:
Comments have been addressed. PR is ready for final review.

bgulanowski:
Needs to be merged with parent branch.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
```diff
@@ -0,0 +1,411 @@
...
+
+@implementation BUYRequestOperationTests
+
+#pragma mark - Setup -
+
+- (void)setUp
+{
+    [super setUp];
+	
+	self.request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:@"https://www.google.com"]];
```
bgulanowski: `www.shopify.com`? 😁

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYRequestOperation.h
```diff
@@ -0,0 +1,49 @@
...
+@protocol BUYSerializable;
+
+typedef void (^BUYRequestOperationCompletion)(NSDictionary * _Nullable json, NSHTTPURLResponse * _Nullable response, NSError * _Nullable error);
+typedef BOOL (^BUYRequestOperationPollingHandler)(NSDictionary * _Nullable json, NSHTTPURLResponse * _Nullable response, NSError * _Nullable error);
+
+@interface BUYRequestOperation : BUYOperation
+
+@property (strong, nonatomic, readonly, nonnull) NSURLSession *session;
+@property (strong, nonatomic, readonly, nonnull) NSURLRequest *originalRequest;
+@property (strong, nonatomic, readonly, nonnull) BUYRequestOperationCompletion completion;
```
bgulanowski: Is this property used by anything outside of this class? Perhaps it should be private.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYRequestOperation.m
```diff
@@ -0,0 +1,175 @@
...
+
+#import "BUYRequestOperation.h"
+#import "BUYSerializable.h"
+
+NSString * const kShopifyError = @"shopify";
+
+typedef void (^BUYRequestJSONCompletion)(NSDictionary *json, NSHTTPURLResponse *response, NSError *error);
+
+#pragma mark - NSURLResponse -
+@interface NSHTTPURLResponse (Conveniece)
```
bgulanowski: typo: "Convenience"

Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
```diff
@@ -0,0 +1,411 @@
...
+		XCTAssertNotNil(response);
+		XCTAssertNil(error);
+		
+		if (response.statusCode == BUYStatusComplete) {
+			[expectation fulfill];
+		}
+		
+		if (response.statusCode == BUYStatusProcessing) {
+			pollCount += 1;
+			if (pollCount == 10) {
```
bgulanowski: Polling once probably fine?

bgulanowski:
Couple more minor comments. Otherwise, 🎩  -- Love it!

PR #169, opened 2016-05-18
Refactor payment provider

gabrieloc:
## What
- Small change to method naming in `<BUYPaymentProviderDelegate>`
- Breaks up checkout error parsing into smaller methods

@bgulanowski @dbart01 @davidmuzi


dbart01:
LGTM :shipit:

PR #170, opened 2016-05-18
Core Data Configurations

### LINKED ISSUE #161: Add Core Data configurations
In order to support Core Data, the SDK framework requires a special build flag, "CORE_DATA_PERSISTENCE", to be enabled. To support embedding the framework in another app, we settled on a solution that determines the inheritance of model classes using this flag. To set the flag correctly in an embedded framework, we must ensure that the build configuration is the same between the parent app and the SDK. So we need custom configurations, ideally named "Debug-core-data" and "Release-core-data", to be added to the SDK that sets this flag.

In an enclosing project, configurations with the same name must exist and be selected, to ensure that Xcode picks the right configuration for the embedded project.

gabrieloc:
**What this does**
Adds new Debug and Release configurations for debugging Core Data, named `Debug-core-data` and `Release-core-data` respectively. These configurations define a macro named `CORE_DATA_PERSISTENCE`, which determine the inheritance of model classes.

Fixes #161 

@dbart01 @bgulanowski @davidmuzi

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK.xcodeproj/project.pbxproj
+++ b/Mobile Buy SDK/Mobile Buy SDK.xcodeproj/project.pbxproj
@@ -1867,6 +1867,308 @@
 			};
 			name = Release;
 		};
+		B2AE98341CED035700FB0C5D /* Debug-core-data */ = {
+			isa = XCBuildConfiguration;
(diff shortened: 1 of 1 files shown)


davidmuzi:
🚢

Issue #159, opened 2016-05-11
Move shipping rates polling into `BUYClient`


bgulanowski:
There is currently logic in the Apple Pay helper for performing polling on the shipping rates endpoint. We need this support for the native checkout, too. Move into `BUYClient` for ease of access.

PR #174, opened 2016-05-19
Add convenience auto-polling API.

dbart01:
### What this does
- implement `BUYCheckoutOperation` to wrap 3 API requests:
 1. Start checkout 
 2. Poll for completion status
 3. Get completed checkout
- converts `completeCheckout:` endpoint to internally begin checkout, poll for completion status and fetch completed checkout via a new `BUYCheckoutOperation`
- converts `getShippingRates:` endpoint to internally poll and return complete shipping rates
- removes `BUYCheckout` from the `storeCreditCard:completion:` callback to be consistent with Android

Fixes: #156 
Fixes: #159 

Depends on #167 


Mobile Buy SDK/Mobile Buy SDK Tests/BUYIntegrationTest.m
```diff
@@ -265,80 +253,35 @@ - (void)completeCheckoutWithToken:(id<BUYPaymentToken>)paymentToken
...
-		XCTAssertNotNil(returnedCheckout);
-		_checkout = returnedCheckout;
-		XCTAssertNotNil(_checkout.order.identifier);
-		XCTAssertNotNil(_checkout.order.statusURL);
-		XCTAssertNotNil(_checkout.order.name);
 		[expectation fulfill];
 		
 		[self confirmCreditCard];
 	}];
-	[self waitForExpectationsWithTimeout:10 handler:^(NSError *error) {
```
davidmuzi: shouldn't be changed

Mobile Buy SDK/Mobile Buy SDK/Data/BUYRouter.m
```diff
@@ -0,0 +1,253 @@
+//
```
davidmuzi: Why is this class here?

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
```diff
@@ -0,0 +1,168 @@
...
+
+#import "BUYCheckoutOperation.h"
+#import "BUYClient+Checkout.h"
+#import "BUYPaymentToken.h"
+#import "BUYCheckout.h"
+#import "BUYRequestOperation.h"
+
+@interface BUYClient (PrivateCheckout)
+
+- (BUYRequestOperation *)beginCheckout:(BUYCheckout *)checkout paymentToken:(id<BUYPaymentToken>)paymentToken completion:(BUYDataCheckoutBlock)block;
```
davidmuzi: Private extensions should be in another header, this is fragile

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
```diff
@@ -0,0 +1,168 @@
...
+		if (weakSelf.isCancelled) {
+			return;
+		}
+		
+		if (!checkout) {
+			[weakSelf finishWithError:error];
+			return;
+		}
+		
+		BUYRequestOperation *pollOperation = [weakSelf.client getCompletionStatusOfCheckoutToken:weakSelf.checkout.token start:NO completion:^(BUYStatus status, NSError *error) {
```
davidmuzi: the nice thing with operations, is that you don't need to nest ops like this, but instead set them as dependancies

davidmuzi:
Some comments.  Should focus on getting #167 first - is this PR is noisy

Mobile Buy SDK/Mobile Buy SDK/Data/BUYRouter.m
dbart01: Sneaked through somehow, will remove.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYIntegrationTest.m
dbart01: Nice catch, wasn't meant to be committed.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
dbart01: In this case the nesting is necessary in order to return a relevant error in case the first or second operation fails. Using dependencies, it makes it impossible to prevent the other operations from running (cancelling them) after an error occurs AND returning that error to the caller. After much experimentation, I believe this is the best solution.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
dbart01: I'm not sure I see how moving them to a separate header is less fragile. The private methods rely on checkout-specific classes and block definitions. Moving them to `BUYClient+Internal.h` for example would require polluting that header with imports for those classes and block definitions.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
davidmuzi: If the callback signatures were similar, or if the BUYOperation also had delegate methods, the error handling for all three operations could be handled by one block/method, since they are all doing the same thing.

It is not impossible to cancel them, you have the references to them.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
davidmuzi: There is an implicit dependancy on this class from BUYClient when doing this.  If the method is changed in BUYClient this class while happily compile while the rug is pulled under it.

dbart01: I agree that's not ideal, but that would remain the case even if the declarations are moved into `BUYClient+Internal.h` or another header. Would love to find a better solution, though.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
dbart01: The problem isn't in handling the error, per se. It's that the operation is considered "finished" upon returning an error, in which case the dependant operations will automatically begin executing despite being completely pointless since the prerequisite operation failed. Either way, it may be technically possible to achieve this using dependencies, but the overhead of additional code and complexity, in my opinion, outweigh the benefit in this case.

davidmuzi: Ok, we can address this in another PR.  

David Muzi

On Fri, May 20, 2016 at 8:38 AM Dima Bart

<
mailto:Dima Bart <notifications@github.com>
> wrote:

In
https://github.com/Shopify/mobile-buy-sdk-ios/pull/174#discussion_r64034186
:

> + > + [super startExecution]; > + > + __weak typeof(self) weakSelf = self; > + > + BUYRequestOperation *beginOperation = [weakSelf.client beginCheckout:weakSelf.checkout paymentToken:weakSelf.token completion:^(BUYCheckout *checkout, NSError *error) { > + if (weakSelf.isCancelled) { > + return; > + } > + > + if (!checkout) { > + [weakSelf finishWithError:error]; > + return; > + } > + > + BUYRequestOperation *pollOperation = [weakSelf.client getCompletionStatusOfCheckoutToken:weakSelf.checkout.token start:NO completion:^(BUYStatus status, NSError *error) {

The problem isn't in handling the error, per se. It's that the operation is considered "finished" upon returning an error, in which case the dependant operations will automatically begin executing despite being completely pointless since the prerequisite operation failed. Either way, it may be technically possible to achieve this using dependencies, but the overhead of additional code and complexity, in my opinion, outweigh the benefit in this case.

—

You are receiving this because you were mentioned.

Reply to this email directly or
https://github.com/Shopify/mobile-buy-sdk-ios/pull/174/files/6508223e3da9d5039fa0b031a5b908ecaf22bb38#r64034186
https://github.com/Shopify/mobile-buy-sdk-ios/pull/174/files/6508223e3da9d5039fa0b031a5b908ecaf22bb38#r64034186

dbart01:
Ready for final review 😋

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
bgulanowski: Can we at least extract the creation of the follow-on operations to separate methods? This method/block has a lot of indenting and is hard to follow.

Is there a clean and simple way to make dependent operations automatically cancel if the preceding operation failed?

bgulanowski: Something that concerns me here: the completion blocks of one operation are referring to the original operation. It looks like most of this code would not execute, because the original operation is going to get removed from the operation queue. It's implicit that whoever creates the first operation has to retain a strong reference to it, otherwise it will be deallocated and the subsequent operations will not happen. Am I wrong here? I think we might have to avoid using references to the operations in the completion blocks. For example, by declaring strong local variables, or perhaps by using properties of the subsequent operations instead of the originating one.

davidmuzi: The only glue that binds the operations together is the checkout object, which should be strongly held by this class.  

This does bring up another point, however, of the confusion of having the operations require an checkout object as an input, and returning another checkout object in its response.  It would be clearer if only the checkout token was required to instantiate each operation (extract the `free` check elsewhere), which should make it clearer that the response is a new object.

As for this implementation, each of the operations could be enqueued with the checkout token (which would not change) marking each as a dependancy on the next.  The problem is if a dependancy is canceled, the cancellation does not propagate to the dependants.  A GroupOperation class could be implemented which incapsulates the necessary logic for this to work; setting up the dependancies and propagating cancellations.

dbart01:
> It looks like most of this code would not execute, because the original operation is going to get removed from the operation queue

The operation's completion block is called before its removed from the queue. This can be verified in the tests.

> It's implicit that whoever creates the first operation has to retain a strong reference to it, otherwise it will be deallocated and the subsequent operations will not happen. Am I wrong here?

The operation queue retains the operation. Also, the `BUYCheckoutOperation` retains a reference to the currently running operation to facilitate cancellation.

In my opinion, chaining operations via callbacks is an efficient and easy-to-understand approach. I think the main problem we're encountering here is that the operation `isFinished` state kicks off subsequent dependencies and does not distinguish between a successful `isFinished` and failed `isFinished` states. The simplest way to handle this without adding complexity, is to chain operations via the callbacks, in my opinion.

dbart01:
> It would be clearer if only the checkout token was required to instantiate each operation

This is a good point and something @sav007 brought up on Android as well. If we go with this approach, it may be necessary to rethink other API that accept complete objects when only a `token` or `identifier` is required.

> The problem is if a dependancy is canceled, the cancellation does not propagate to the dependants.

The dependant operations won't execute if the their prerequisite operation is cancelled. We don't call the completion on cancelled operations, but how do we return the relevant `NSError` immediately from any one _failed_ operation?

>  A GroupOperation class could be implemented which incapsulates the necessary logic for this to work

Good point, I'll explore this solution further.

davidmuzi:
> The dependant operations won't execute if the their prerequisite operation is cancelled.
Not true

dbart01:
That's right, my mistake.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYRequestOperation.m
```diff
@@ -144,9 +143,16 @@ - (NSURLSessionDataTask *)requestUsingPollingIfNeeded:(NSURLRequest *)request co
 		 * the polling process.
 		 */
 		if (self.pollingHandler && self.pollingHandler(json, response, error)) {
-			NSURLSessionDataTask *task = [self requestUsingPollingIfNeeded:request completion:completion];
-			self.runningTask = task;
-			[task resume];
+
+			dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(self.pollingInterval * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+				if (self.isCancelled) {
```
bgulanowski: You're using the method name and the property syntax. Sure, it works for the getter, but try this with the setter and you'll regret it. (OK it doesn't apply here with a read-only property, but…) We should be consistent and use `cancelled` with property dot syntax.

bgulanowski:
👍  looks great!

PR #175, opened 2016-05-19
Clean up BUYApplePayHelpers

### LINKED ISSUE #154: Clean up ApplePayHelpers
ApplePayHelpers has accumulated a lot of old code for backward compatibility reasons to v1.

- remove network code - require `BUYShop` to be passed in
- remove redundant PKPaymentController delegate methods
- extract polling

gabrieloc:
**What this does**
- Makes `shop` a required argument for initialization
- Removes redundant payment authorization methods
- Polling work handled by https://github.com/Shopify/mobile-buy-sdk-ios/pull/174

Fixes #154 

@davidmuzi @dbart01

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
@@ -1,5 +1,5 @@
 //
-//  BUYApplePayPaymentControllerDelegate.m
+//  BUYApplePayAuthorizationDelegate.m
 //  Mobile Buy SDK
 //
 //  Created by Shopify.
@@ -24,7 +24,7 @@
 //  THE SOFTWARE.
 //
 
-#import "BUYApplePayPaymentControllerDelegate.h"
+#import "BUYApplePayAuthorizationDelegate.h"
 #import "BUYApplePayAdditions.h"
 #import "BUYApplePayToken.h"
 #import "BUYClient+Checkout.h"
@@ -37,20 +37,20 @@
 
 const NSTimeInterval PollDelay = 0.5;
 
-@interface BUYApplePayPaymentControllerDelegate ()
+@interface BUYApplePayAuthorizationDelegate ()
 @property (nonatomic, strong) BUYCheckout *checkout;
 @property (nonatomic, strong) BUYClient *client;
 
 @property (nonatomic, strong) NSArray *shippingRates;
 @property (nonatomic, strong) NSError *lastError;
 
-@property (nonatomic, strong) BUYShop *shop;
+@property (nonatomic, strong) NSString *shopName;
 
 @end
 
-@implementation BUYApplePayPaymentControllerDelegate
+@implementation BUYApplePayAuthorizationDelegate
 
-- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout shop:(BUYShop *)shop
+- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout shopName:(NSString *)shopName
 {
 	NSParameterAssert(client);
 	NSParameterAssert(checkout);
@@ -60,7 +60,7 @@ const NSTimeInterval PollDelay = 0.5;
 	if (self) {
 		self.client = client;
 		self.checkout = checkout;
-		self.shop = shop;
+		self.shopName = shopName;
 	}
 	
 	return self;
--- a/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
@@ -39,8 +39,6 @@ const NSTimeInterval PollDelay = 0.5;
 
 @interface BUYApplePayAuthorizationDelegate ()
 
-@property (nonatomic, strong) BUYCheckout *checkout;
-
 @property (nonatomic, strong) NSArray *shippingRates;
 @property (nonatomic, strong) NSError *lastError;
 
@@ -150,7 +148,7 @@ const NSTimeInterval PollDelay = 0.5;
 	
 	[self.client updateCheckout:self.checkout completion:^(BUYCheckout *checkout, NSError *error) {
 		if (checkout && error == nil) {
-			self.checkout = checkout;
+			_checkout = checkout;
 		}
 		else {
 			self.lastError = error;
@@ -174,7 +172,7 @@ const NSTimeInterval PollDelay = 0.5;
 		
 		[self.client updateCheckout:self.checkout completion:^(BUYCheckout *checkout, NSError *error) {
 			if (checkout && error == nil) {
-				self.checkout = checkout;
+				_checkout = checkout;
 				[self getShippingRates:self.checkout completion:completion];
 			}
 			else {
@@ -218,7 +216,7 @@ const NSTimeInterval PollDelay = 0.5;
 			if ([shippingMethods count] > 0) {
 				[self selectShippingMethod:shippingMethods[0] completion:^(BUYCheckout *checkout, NSError *error) {
 					if (checkout && error == nil) {
-						self.checkout = checkout;
+						_checkout = checkout;
 					}
 					completion(error ? PKPaymentAuthorizationStatusFailure : PKPaymentAuthorizationStatusSuccess, shippingMethods, [self.checkout buy_summaryItemsWithShopName:self.shopName]);
 				}];
--- a/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
@@ -40,6 +40,8 @@ const NSTimeInterval PollDelay = 0.5;
 
 @interface BUYApplePayAuthorizationDelegate ()
 
+@property (nonatomic, strong) BUYCheckout *checkout;
+
 @property (nonatomic, strong) NSArray *shippingRates;
 @property (nonatomic, strong) NSError *lastError;
 
@@ -92,14 +94,14 @@ const NSTimeInterval PollDelay = 0.5;
 	
 	[self.client updateCheckout:self.checkout completion:^(BUYCheckout *checkout, NSError *error) {
 		if (checkout && error == nil) {
-			_checkout = checkout;
+			self.checkout = checkout;
 			
 			id<BUYPaymentToken> token = [[BUYApplePayToken alloc] initWithPaymentToken:payment.token];
 			
 			//Now that the checkout is up to date, call complete.
 			[self.client completeCheckout:checkout paymentToken:token completion:^(BUYCheckout *checkout, NSError *error) {
 				if (checkout && error == nil) {
-					_checkout = checkout;
+					self.checkout = checkout;
 					
 					[self pollUntilCheckoutIsComplete:self.checkout completion:completion];
 				}
@@ -150,7 +152,7 @@ const NSTimeInterval PollDelay = 0.5;
 	
 	[self.client updateCheckout:self.checkout completion:^(BUYCheckout *checkout, NSError *error) {
 		if (checkout && error == nil) {
-			_checkout = checkout;
+			self.checkout = checkout;
 		}
 		else {
 			self.lastError = error;
@@ -174,7 +176,7 @@ const NSTimeInterval PollDelay = 0.5;
 		
 		[self.client updateCheckout:self.checkout completion:^(BUYCheckout *checkout, NSError *error) {
 			if (checkout && error == nil) {
-				_checkout = checkout;
+				self.checkout = checkout;
 				[self getShippingRates:self.checkout completion:completion];
 			}
 			else {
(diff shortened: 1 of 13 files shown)


Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
```diff
@@ -132,3 +93,5 @@
 - (BUYAddress *)buyAddressWithContact:(PKContact *)contact NS_AVAILABLE_IOS(9_0);
```
davidmuzi: this category should be put in another file

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
```diff
@@ -34,17 +34,9 @@
...
-/**
- *  Initializes a helper to support Apple Pay
- *
- *  @param client   A configured client
- *  @param checkout The checkout which is to be completed using Apple Pay
- *
- *  @return helper object
- */
-- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout;
+@interface BUYApplePayHelpers : NSObject <PKPaymentAuthorizationViewControllerDelegate>
```
davidmuzi: now that this class has a single responsibility, lets rename to BUYApplePayPaymentControllerDelegate, or similar

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
```diff
@@ -55,39 +47,8 @@
...
-- (void)updateCheckoutWithAddress:(ABRecordRef)address completion:(void (^)(PKPaymentAuthorizationStatus status, NSArray *shippingMethods, NSArray *summaryItems))completion NS_DEPRECATED_IOS(8_0, 9_0, "Use the CNContact backed `updateCheckoutWithContact:completion:` instead");
-
-/**
- *  Call this method in the PKPaymentAuthorizationViewControllerDelegate `paymentAuthorizationViewController:didSelectShippingAddress:completion`
- *
- *  @param contact    The selected contact
- *  @param completion the completion block called after the shipping address is updated on the checkout
- */
-- (void)updateCheckoutWithContact:(PKContact*)contact completion:(void (^)(PKPaymentAuthorizationStatus, NSArray *shippingMethods, NSArray *summaryItems))completion NS_DEPRECATED_IOS(8_0, 9_0, "Use `PKPaymentAuthorizationViewControllerDelegate` instead");
+- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout shop:(BUYShop *)shop NS_DESIGNATED_INITIALIZER;
```
davidmuzi: we can just pass in the shopName

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
```diff
@@ -66,25 +61,12 @@ - (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)check
...
-		}
-		else {
-			[self.client getShop:^(BUYShop *shop, NSError *error) {
-				
-				if (shop) {
-					self.shop = shop;
-				}
-			}];
-		}
+		self.shop = shop;
```
davidmuzi: should assert on shop(name)

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
dbart01: How about `BUYPaymentControllerDelegate`?

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
dbart01: Should use `BUYAssert` instead of `NSParameterAssert` for both assertions.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
```diff
@@ -51,11 +51,6 @@ @interface BUYApplePayHelpers ()
 
 @implementation BUYApplePayHelpers
 
-- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout
-{
-	return [self initWithClient:client checkout:checkout shop:nil];
-}
-
 - (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout shop:(BUYShop *)shop
```
dbart01: Can we make these properties (client, checkout, shop) `readonly` and assign them via instance variable in the initializer? Example: `_shop = shop`.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
dbart01: This would also remove the need for `#import <Buy/BUYModelManager.h>`

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
davidmuzi: This class is specifically for Apple Pay, I'd prefer to make that evident from the class name

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.h
dbart01: Where would the shop name be used? `getShop:` takes only the completion as a parameter.

davidmuzi: This class only needs the shopName to do its job, not the entire BUYShop object

dbart01: True, but the user would have to fetch the shop anyway in order to pass in the `shopName`, in which case they'll have the complete `BUYShop` object. Just a thought.

davidmuzi: In which case they can just use shop.shopName.  Requiring the entire object imposes another restriction on the API consumer, and prevents them from sending a hard-coded string, which they would likely have on a one-shop app

dbart01: Actually, looking into this further, `shop` or `shopName` isn't a requirement. It's only used in `-[BUYCheckout buy_summaryItemsWithShopName:]` where the `shopName` is `nullable` and uses "TOTAL" if the `shopName` isn't provided. This means that the initializer should require `shopName` at all. Instead, it can be a property on this helper class:
```objc
@property (strong, nonatomic, nullable) NSString *shopName;
```

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
gabrieloc: No longer necessary if passing in the name

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
gabrieloc: Looks like we need to set `checkout` a few times, but the other properties can be made readonly.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYModelManager+ApplePay.h
```diff
@@ -0,0 +1,33 @@
+//
+//  BUYModelManager+ApplePay.h
+//  Mobile Buy SDK
+//
+//  Created by Gabriel O'Flaherty-Chan on 2016-05-20.
+//  Copyright © 2016 Shopify Inc. All rights reserved.
+//
+
+#import <Buy/BUYModelManager.h>
+@class BUYAddress;
```
dbart01: Nit: Can we not mix forward declarations with import statements? `@class` should be after all imports.

gabrieloc:
Comments have been addressed 📡

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
dbart01: We can still set the checkout via instance variable. But it shouldn't be mutable from the outside.

gabrieloc: What's the harm in using a setter though?

dbart01: Its not strictly necessary but I prefer to only expose mutable properties when they are required. As far as I can tell, there's never a need to set the `BUYCheckout` from the outside after initialization.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -24,67 +24,44 @@
...
-{
-	return [self initWithClient:client checkout:checkout shop:nil];
-}
+@implementation BUYApplePayAuthorizationDelegate
 
-- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout shop:(BUYShop *)shop
+- (instancetype)initWithClient:(BUYClient *)client checkout:(BUYCheckout *)checkout shopName:(NSString *)shopName
 {
 	NSParameterAssert(client);
 	NSParameterAssert(checkout);
```
davidmuzi: assert on shopName

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -264,14 +216,14 @@ - (void)getShippingRates:(BUYCheckout *)checkout completion:(void (^)(PKPaymentA
 			if ([shippingMethods count] > 0) {
 				[self selectShippingMethod:shippingMethods[0] completion:^(BUYCheckout *checkout, NSError *error) {
 					if (checkout && error == nil) {
-						self.checkout = checkout;
+						_checkout = checkout;
```
bgulanowski: If we're going to standardize on using accessors for reading properties, we should use accessors for writing them. We should define private readwrite property for `checkout`.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYModelManager+ApplePay.h
```diff
@@ -0,0 +1,34 @@
+//
+//  BUYModelManager+ApplePay.h
+//  Mobile Buy SDK
+//
+//  Created by Gabriel O'Flaherty-Chan on 2016-05-20.
```
bgulanowski: Copyright message.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
gabrieloc: Can we rely on nullability?

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
gabrieloc: Check out https://github.com/Shopify/mobile-buy-sdk-ios/pull/175#discussion_r64053758

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
dbart01: If "warnings as errors" are disabled nullability will simply be a warning and isn't enough, in my opinion.

gabrieloc: That's fair

dbart01: Can we use `BUYAssert` macro instead of `NSParameterAssert`?

gabrieloc:
@davidmuzi @dbart01 Had to update the stubbed test shop response with the shop name after adding that parameter assertion. Should this info be updated anywhere else as well?

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayHelpers.m
bgulanowski: Whether to use ivar or property notation for private read/write properties is a convention. However, the standard practise is moving towards the latter (property/setter) access. I've personally flip-flopped on this. I do believe that ivar access is correct for `-init` methods, but we should use setters/property access elsewhere. There are a few reasons, including KVO.

davidmuzi: Yes, lets stick with ivars in init, and use properties elsewhere.  You can redefine private properties in a private extension with readwrite so they can be set in the implementation file

gabrieloc: I've redeclared this property in a private extension.

dbart01:
LGTM 👍

PR #183, opened 2016-05-25
Fix payment provider tests

### LINKED ISSUE #182: Payment provider tests causing failures
There seems to be a bug with `SFSafariViewController`. The payment provider uses the safari controller in this test: `-(void)testWebPresentationCallbacks`, which is causing a crash resulting in failed tests on CI.

The solution is to mock / inject a fake `SFSafariViewController` when testing.

dbart01:
### What this does
- adds `BUYFakeSafariController` to inject into web payment provider during testing
- adds initializer to web payment provider to accept injected fake safari controller
- minor cleanup and reorganization

Fixes #182 


Mobile Buy SDK/Mobile Buy SDK Tests/BUYFakeSafariController.m
```diff
@@ -0,0 +1,60 @@
...
+
+#pragma mark - Fake Controller -
+
+@interface BUYFakeSafariController ()
+
+@end
+
+@implementation BUYFakeSafariController
+
+- (instancetype)initWithURL:(NSURL *)URL {
```
davidmuzi: Why do we need this implementation?  Not only don't we need to use this constructor (even though it is the designated one), but why do we need to have "readerMode" enabled?

Mobile Buy SDK/Mobile Buy SDK Tests/BUYFakeSafariController.m
```diff
@@ -0,0 +1,60 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+#import "BUYFakeSafariController.h"
+
+#pragma mark - Delegator -
+
+@interface BUYFakeDelegator : NSObject
```
davidmuzi: BUYMock* ?

davidmuzi:
Couple comments.  Need merging from integration branch.  Good use of DI!

Mobile Buy SDK/Mobile Buy SDK Tests/BUYFakeSafariController.m
dbart01: The init methods exist on the fake safari controller because they mimic the real safari controller. The "readerMode" was set to `YES` for no reason, just a coin flip between the two values. For consistency, I included both initializers that are available on `SFSafariViewController`.

dbart01:
After some discussion with @bgulanowski, we decided on another approach that would avoid exposing the initializer with `webClass` parameter.

bgulanowski:
:shipit:

PR #184, opened 2016-05-25
Remove duplicate references in target.

bgulanowski:
# What

This is a simple project cleanup change.

# Why

There were two references to a header in both framework targets, and you don't want that. Probably caused by merge.

@dbart01 @gabrieloc


dbart01:
Needs to resolve merge conflict.

dbart01:
🚢

PR #185, opened 2016-05-25
Add customer logout and update API.

### LINKED ISSUE #179: Add missing customer API
We are missing functionality in the customer API:

- logout
- update

If log out fails because the token doesn't exist, we shouldn't report that as an error. If it succeeds, the `BUYClient` should automatically delete it's copy of the token.

@dbart01 @gabrieloc

dbart01:
#### What this does
- adds API for updating customer info
- adds API for logging out customer (deleting token on the server)
- adds tests for above functions
- adds stub for logout API call
- renames misleading `customerToken` -> `token`. This is **NOT** a customer auth token that we pass in the header.

Fixes #179 
Depends on #174 

@bgulanowski @gabrieloc @davidmuzi


Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest_Customer.m
```diff
@@ -153,4 +142,94 @@ - (void)testCustomerLogin
...
+	}];
+	
+	XCTestExpectation *expectation = [self expectationWithDescription:NSStringFromSelector(_cmd)];
+	
+	BUYAccountCredentialItem *email    = [BUYAccountCredentialItem itemWithEmail:self.customerEmail];
+	BUYAccountCredentials *credentials = [BUYAccountCredentials credentialsWithItems:@[email]];
+	
+	[self.client loginCustomerWithCredentials:[self credentialsForLogin] callback:^(BUYCustomer * _Nullable customer, NSString * _Nullable token, NSError * _Nullable error) {
+		
+		[OHHTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest * _Nonnull request) {
```
bgulanowski: Can this stub request be added earlier than here?

I don't like how all these mocking code chunks noisy up the core test logic.

I wonder if it would be better, in the long run, to install all stubs in `+setup`, and define specific tests per URL, instead of using `-shouldUseMocks` in the test (but only install the stubs if `-shouldUseMocks` is `YES`.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.h
```diff
@@ -93,27 +97,44 @@ typedef void (^BUYDataOrdersBlock)(NSArray <BUYOrder*> * _Nullable orders, NSErr
 - (BUYRequestOperation *)createCustomerWithCredentials:(BUYAccountCredentials *)credentials callback:(BUYDataCustomerTokenBlock)block;
 
 /**
- *  POST /api/customers/customer_token
- *  Logs in an existing customer
- *  Expects email and password
+ *  PUT /api/customers/:customer_id/activate
```
bgulanowski: One day, git will be integrated with the language interpreter. It will realize that changes that only move code around—without changing its meaning (like moving method/function declarations/implementations)—won't create these diffs.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.m
```diff
@@ -63,64 +65,64 @@ - (BUYRequestOperation *)createCustomerWithCredentials:(BUYAccountCredentials *)
...
 			block(nil, nil, error);
 		}
 	}];
 }
 
-- (BUYRequestOperation *)loginCustomerWithCredentials:(BUYAccountCredentials *)credentials callback:(BUYDataCustomerTokenBlock)block
+- (BUYRequestOperation *)updateCustomerWithCredentials:(BUYAccountCredentials *)credentials customerID:(NSString *)customerID callback:(BUYDataCustomerBlock)block
 {
-	return [self createTokenForCustomerWithCredentials:credentials customerJSON:nil callback:block];
+	NSURL *route = [self urlForCustomersWithID:customerID];
```
bgulanowski: Can we call urls `url` instead of `route`?

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.h
dbart01: Looking forward to that day 👍

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTest_Customer.m
dbart01: Agreed. This is something I've been dreaming about for a while now. Not sure if this is something worth doing at this point in time, however. Might be best suited for a 2.0.1 release.

dbart01:
👍  ?

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.m
```diff
@@ -63,64 +65,64 @@ - (BUYRequestOperation *)createCustomerWithCredentials:(BUYAccountCredentials *)
...
+			[self loginCustomerWithCredentials:[credentials credentialsByAddingItems:@[emailItem]] callback:block];
 		}
 		else {
 			block(nil, nil, error);
 		}
 	}];
 }
 
-- (BUYRequestOperation *)loginCustomerWithCredentials:(BUYAccountCredentials *)credentials callback:(BUYDataCustomerTokenBlock)block
+- (BUYRequestOperation *)updateCustomerWithCredentials:(BUYAccountCredentials *)credentials customerID:(NSString *)customerID callback:(BUYDataCustomerBlock)block
```
bgulanowski: If I understand this, this allows the customer to change their first name, last name, email, or password. Seems reasonable, although odd to use credentials instead of just using the customer object. But I suppose since most fields on customer are not editable, this is sensible.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYRequestOperation.m
```diff
@@ -169,14 +168,24 @@ - (NSURLSessionDataTask *)request:(NSURLRequest *)request completion:(BUYRequest
 			json = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
 		}
 		
-		NSHTTPURLResponse *httpResponse = (id)response;
-		if (!httpResponse.successful && !error) {
-			error = [[NSError alloc] initWithDomain:kShopifyError code:httpResponse.statusCode userInfo:json];
+		NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+		if (!error) {
```
bgulanowski: This change is a bit confusing. Seems funny to seemingly create an error using a URL response. How about just passing in the status code?

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYRequestOperation.m
```diff
@@ -169,14 +168,24 @@ - (NSURLSessionDataTask *)request:(NSURLRequest *)request completion:(BUYRequest
...
 		}
 		
 		completion(json, httpResponse, error);
 	}];
 	return task;
 }
 
+#pragma mark - Error -
+
+- (NSError *)errorFromJSON:(NSDictionary *)json response:(NSHTTPURLResponse *)response
```
bgulanowski: This would seem less weird if the response code was provided as the first argument.

Although, it also seems odd to wrap the body of a method, even a one-liner, with a conditional. Usually it's the job of the caller to perform conditional checks. Ideally, methods always do what they say they do, not conditionally. I think I liked the original design a bit better.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Customers.m
dbart01: That's correct.

bgulanowski:
Couple small suggestions, otherwise good.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYRequestOperation.m
dbart01: Very true. Somehow missed that I created a category on `NSHTTPURLResponse` to provide a `success` flag based on whether `the statusCode` is a `2xx` or not. Using that within the caller eliminates the need to this method altogether. Will remove.

carrotalan:
Any update on when the log out feature can be implemented into master?

PR #189, opened 2016-05-27
Merchant app integration 

gabrieloc:
**What this does**
- Exposes keys used for customer API
- Changes to `BUYCollection` and `BUYProduct` for normalizing how identifiers are represented
- Exposes update methods in `BUYLineItem` and removes deprecated initializers (which won't work anyways)
- Exposes `BUYClient+CheckoutHelpers.h`
- Change how `BUYLineItem` objects are created in `BUYCheckout` (in response to broken initializers)

@bgulanowski @dbart01 @davidmuzi


Mobile Buy SDK/Mobile Buy SDK/Buy.h
```diff
@@ -65,6 +65,7 @@ FOUNDATION_EXPORT const unsigned char BuyVersionString[];
 #import <Buy/BUYClient.h>
 #import <Buy/BUYClient+Customers.h>
 #import <Buy/BUYClient+Checkout.h>
+#import <Buy/BUYClient+CheckoutHelpers.h>
```
dbart01: What is the need for exposing this category? It has only a single method `- (void)handleCheckout:completion:`, which in my opinion is really vague and confusing. If we expose this as public API than there's really no need for this to be in a separate file / category, and I think it should be renamed to something that describes its purpose better.

bgulanowski: Used to only exist in Merchant App, so yeah, maybe we should do something nicer. If we are going to improve it, then `-handleCheckout:` is also crappy name which should instead be `-updateOrCreateCheckout:completion:`.

gabrieloc: I think it makes sense to rename and move to `BUYCheckout`

bgulanowski:
This looks good to me, although tests are failing on CI.

bgulanowski:
I think this is fine to merge after a rebase.

gabrieloc:
Merging on 🍵

PR #192, opened 2016-05-30
Add separate network request queue to avoid deadlock. Minor refactor.

### LINKED ISSUE #190: Fix possible deadlock
When scheduling `BUYRequestOperation`s and network requests on the same queue, it possible that they'll wait on each other indefinitely if the queue doesn't support enough concurrent operations.

Solution is to use two separate queue for operations and network requests.

dbart01:
### What this does
- adds a separate, private network queue for `NSURLSessionDataTask` requests
- minor refactoring and cleanup

Fixes #190 

@bgulanowski @davidmuzi

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.h
@@ -96,11 +96,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nonnull) BUYModelManager *modelManager;
 
 /**
- *  Queue on which all request operation are executed
- */
-@property (nonatomic, strong, readonly, nonnull) NSOperationQueue *requestQueue;
-
-/**
  *  Queue on which network completion callbacks will be executed
  */
 @property (nonatomic, strong, nonnull) NSOperationQueue *callbackQueue;
--- a/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient.m
@@ -34,7 +34,7 @@ static NSString * const BUYClientJSONMimeType = @"application/json";
 @interface BUYClient () <NSURLSessionDelegate>
 
 @property (nonatomic, strong) NSURLSession *session;
-@property (nonatomic, strong) NSOperationQueue *networkQueue;
+@property (nonatomic, strong) NSOperationQueue *requestQueue;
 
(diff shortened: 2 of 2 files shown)


davidmuzi:
The queue used for the NSURLSession is not a queue for the network request to execute on, its the queue where the callbacks are called from.  This queue is intended to be public so the caller can have them scheduled on the queue of their choosing.

> An operation queue for scheduling the delegate calls and completion handlers. The queue need not be a serial queue. If nil, the session creates a serial operation queue for performing all delegate method calls and completion handler calls.

dbart01:
The `NSURLSession`'s `delegateQueue` is used for callbacks, that's true. However, it's also used for network requests' delegate method calls of the `NSURLSession`. Since `NSURLSession` is now an internal implementation detail, I feel that the `delegateQueue` should also remain internal. In addition, the `delegateQueue` cannot be set after `NSURLSession` initialization. The user can set the `requestQueue` to whatever they and that will ensure that callbacks are executed on that queue.

dbart01:
> Why not pass in nil?
We could. In that case `NSURLSession` will default to a serial queue. I thought the SDK might benefit from an explicit concurrent queue. The decision is up for debate.

bgulanowski:
The naming is confusing. And I think the extra queue is unnecessary.

There are two levels of callback: the client's, and the app's. The client's callbacks should run in the background on a private, concurrent queue. I don't see why that queue is exposed. If that queue is concurrent, request operations should be able to run there, too. Where is the risk of deadlocks coming from?

The app's callbacks should happen on whatever queue the app specifies, by default the main queue. That is the queue that should be exposed.

dbart01:
The queues are not just used for callbacks. We need three queues:
- queue owned and managed by `NSURLSession` for handling delegate callbacks AND completion callback
- queue that will enqueue `BUYRequestOperation`s. This should be a different queue than the `NSURLSession` queue to avoid a deadlock situation where the operation kicks off a request scheduled on the same queue as the operation. In that case, the callback from the session will be waiting on the operation, which is waiting on the callback.
- finally, we have a "convenience" queue that schedules user-level callbacks, by default on the main thread

On second though, we don't need to expose the first two queue. Just the last one.

davidmuzi:
In that case, you would need one private queue to enqueue the BUYRequestOperations, a public queue for callbacks which defaults to mainQueue, and pass in nil for the NSURLSession callbacks, and have that use its own serial queue - there is no need for concurrency.

dbart01:
That's fine with me.

PR #205, opened 2016-06-02
Add missing product handle API.

dbart01:
### What this does
- adds missing API that allows fetching a product by its unique `handle`
- adds integration test for product handle API

Fixes #194 
Relies on #204

@bgulanowski @davidmuzi

PR #207, opened 2016-06-02
Prevent exposing BUYRequestOperation

### LINKED ISSUE #206: Avoid exposing internal NSOperation subclasses in the BUYClient
The `BUYClient` should return an `NSOperation` handle to each executed operation instead of specific subclass, such as `BUYCheckoutOperation` or `BUYRequestOperation`.

dbart01:
### What this does
- hides `BUYRequestOperation` and all other operation subclasses from public interface
- all API requests now return `NSOperation`
- removes duplicate tests for product listings
- adds test for sort order conversion
- renames `Data` -> `Client` group in Xcode hierarchy to more accurately reflect its contents

Fixes #206 
Depends on #205 

@bgulanowski @davidmuzi


dbart01:
Ready for review. It looks like having `nil` handlers for `waitForExpectationsWithTimeout:` may be causing CI test failures. Having briefly tested on a couple commits, it looks like it might have solved the issue. Hopefully that's the case.

bgulanowski:
🚢  it.

Issue #176, opened 2016-05-19
Sample Apps


davidmuzi:
Currently there are 4 sample apps:
- Advanced Sample App
  - product single, apple pay, products and collections listing
- Swift Sample
  - Swift interoperability 
- Customer Sample
  - demonstrates customer APIs
  - Swift
- Web Sample
  - demonstrates using a web view with Apple Pay

### Discussion

The web sample goes against the spirit of the SDK, and always seemed like a workaround to get Apple Pay in a "native" app. Id vote to remove this sample app.

Since the customer sample app is written in swift, the swift sample doesn't have a purpose any longer.  For clarity, we should append `(Swift)` or `(Obj-C)` next to the folders for each sample app.

Remaining question, should we add the customer sample into the advanced sample app?

@fananta @bgulanowski

fananta:
As I understand it the Advanced Sample should provide an example of utilizing major components of the SDK. If that's the case then I would vote yes, although not a big deal if we leave it as is in the Customer Sample (Swift).

Thoughts:
- remove _Web Sample App_ (signals our developers that our focus is on native)
- customer API example also in _Advanced Sample App_

dbart01:
I'm favour of removing the Web sample app as it provides very little "sample" of the SDK. Regarding the other three apps, I would propose something a little different. Currently, we have no real way of testing Swift interoperability (whether it's Unit Tests written in Swift or sample apps) that cover most of the SDK functionality. Only the Customer sample and the Swift sample (covers very little) exist for that. My proposal, then, is to continue building the Customer sample into a Swift version of the advanced sample to cover all major functionality. Not only will it provide a platform for us to test on real devices, it would also provide great insight into the downfalls that exist with Swift interoperability. This full-featured sample would eventually replace all other sample apps resulting in less clutter.

davidmuzi:
Having swift unit tests would be the best approach for consuming all the APIs and much easier to develop and maintain than another sample app. Let's create a ticket for swift unit tests. 

David Muzi

On Fri, May 20, 2016 at 8:24 AM Dima Bart

<
mailto:Dima Bart <notifications@github.com>
> wrote:

[...]

PR #208, opened 2016-06-02
Fixes for Swift Sample app

bgulanowski:
# What

Update the Swift Sample App.

# Why

Swift Sample did not compile after 2.0 SDK API and design chagnes. Partial fix for #176 

# How

Update to simplify some API and use a model manager, as prescribed.

@dbart01 @davidmuzi


dbart01:
🚢  :shipit: 👍

PR #209, opened 2016-06-03
Remove Buy Static target.

### LINKED ISSUE #178: Remove Universal Static Framework target.
It looks like this target is no longer necessary. It may be used by CocoaPods so need confirmation that it's not longer used before removing.

dbart01:
### What this does
- updates universal aggregate target to build dynamic target with static flag
- organizes scheme order to match the target order
- refactors the universal build script

Fixes #178 

@davidmuzi @bgulanowski

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK.xcodeproj/project.pbxproj
+++ b/Mobile Buy SDK/Mobile Buy SDK.xcodeproj/project.pbxproj
@@ -33,41 +33,23 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


bgulanowski:
This appears to be putting the intermediates into the build folder, instead of the intermediates folder. On my machine, the "Products" folder contains "Mobile Buy SDK.build" along with "Debug-universal", but the "Mobile Buy SDK.build" folder should be in the "Intermediates" folder (beside the "Products" folder).

bgulanowski:
- [ ] Clean build action does not clean anything.

PR #211, opened 2016-06-03
Fix equality test in BUYOptionValue

bgulanowski:
# What

This changes the test for equality in BUYOptionValue.

# Why

Without it, different values for the same option look equal, meaning some options will not show up with added to a set. A consequence was that in the Advanced Sample, all products show only one option.

# How

Use the `value`, not the `optionId` (which is basically equivalent to using the `name`, since both refer to properties on the related `BUYOption`).

@dbart01 @davidmuzi @gabo


dbart01:
🚢

PR #215, opened 2016-06-06
Add typed collection for Swift support.

dbart01:
### What this does
Using `CoreData` models required the use of `NSOrderedSet` for certain ordered collections. Since `NSOrderedSet` doesn't have a typed equivalent in Swift, none of the type information transfers over to Swift, making it very inconvenient to use those properties. Here, what we do is:
- add typed collections for `NSOrderedSet` properties

@bgulanowski @davidmuzi

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYCart.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYCart.h
@@ -27,8 +27,16 @@
 #import "_BUYCart.h"
 
 @class BUYProductVariant;
+@class BUYLineItem;
 
 @interface BUYCart : _BUYCart {}
+
+/**
+ * Array of BUYCartLineItem objects in the cart
+ *
+ * These are different from BUYLineItem objects. The line item objects do include the BUYProductVariant.
+ */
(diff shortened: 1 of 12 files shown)


dbart01:
Ping!

bgulanowski:
I don't understand why we should be declaring these arrays. The relationships are sets, what's wrong with those?

dbart01:
In Swift, these collections are untyped. While means that we get something like this:
```
let lineItems: AnyObject = checkout.lineItems
```
This very inconvenient to use since the solution is to always force-cast to the desired type (which requires a trip to the doc, or the header):
```
let lineItems: [BUYLineItem] = checkout.lineItems as! [BUYLineItem]
```

bgulanowski:
But they should not be _arrays_, they should be sets. If we want to annotate the relationships, we should do that via the mogenerator templates.

dbart01:
They can't be typed sets. That's the problem. Swift doesn't transfer light-weight generics annotation of `NSOrderedSet` because an ordered set doesn't exist in Swift. The alternative is to use a `NSSet<Type>` but in that case we loose order information.

bgulanowski:
OK, this is straightforward. Looks good. 👍

PR #219, opened 2016-06-07
Add nullability annotations

dbart01:
### What this does
Swift like nullability annotations. This adds them, and also organizes various things.

@bgulanowski @davidmuzi


Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.h
```diff
@@ -25,8 +25,8 @@
 //
 
 #import <CoreData/CoreData.h>
-
-#import <Buy/BUYObject.h>
+#import "BUYObject.h"
```
bgulanowski: Why did you change the import style here? This form may require additional header search paths, instead of relying on the framework implicit search path.

bgulanowski:
There are too many different kinds of changes in this PR. We need PRs to stay focussed. Please break it up.

The import directives in headers in frameworks must use the framework style, because they are consumed by clients that must also use that style.

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.h
dbart01: For consistency. We always use local-file `#import` style aside from several cases (such as this one). Using the framework-style `#import` also requires that the header is `public` and exposed in the umbrella header, which we may or may not want to do for some files. Using the quotation mark import requires no changes if we decided to go either way.

dbart01:
Correct me if I'm wrong but all this does is reorder some Xcode groups and adds nullability annotations.

bgulanowski:
This PR does three things:

1. Changing import directives (which I'm still against)
2. moving files around in the project organizer
3. modifies interfaces

Would really prefer separate PRs.

bgulanowski:
Headers that are exposed to clients of a framework must use framework import style. We should not require framework clients to add search paths.

dbart01:
Closing in favour of #224 & #225

PR #220, opened 2016-06-07
Add internal API for aggregate checkout status operation

dbart01:
### What this does
- add `BUYAggregateOperation` as a generic operation sequencer
- add `BUYStatusOperation` to poll completion status and get checkout
- use `BUYStatusOperation` as replacement for two operations in `BUYCheckoutOperation`
- expose `BUYStatusOperation` in internal header

Depends on #224 and #225 

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.m
@@ -29,6 +29,7 @@
 #import "BUYClient+Routing.h"
 #import "BUYRequestOperation.h"
 #import "BUYCheckoutOperation.h"
+#import "BUYStatusOperation.h"
 #import "BUYAddress.h"
 #import "BUYCheckout.h"
 #import "BUYGiftCard.h"
@@ -104,7 +105,7 @@
 
 - (BUYOperation *)completeCheckoutWithToken:(NSString *)checkoutToken paymentToken:(id<BUYPaymentToken>)paymentToken completion:(BUYDataCheckoutBlock)block
 {
-	BUYCheckoutOperation *operation = [[BUYCheckoutOperation alloc] initWithClient:self checkoutToken:checkoutToken token:paymentToken completion:block];
+	BUYCheckoutOperation *operation = [BUYCheckoutOperation operationWithClient:self checkoutToken:checkoutToken token:paymentToken completion:block];
 	[self startOperation:operation];
 	return operation;
 }
@@ -150,6 +151,15 @@
 	return [self getRequestForURL:url start:start completionHandler:^(NSDictionary *json, NSHTTPURLResponse *response, NSError *error) {
 		block([self statusForStatusCode:response.statusCode error:error], error);
 	}];
+}
+
+- (NSOperation *)pollCompletionStatusAndGetCheckoutWithToken:(NSString *)token start:(BOOL)start completion:(BUYDataCheckoutBlock)block
+{
+	BUYStatusOperation *operation = [BUYStatusOperation operationWithClient:self checkoutToken:token completion:block];
+	if (start) {
+		[self startOperation:operation];
+	}
+	return operation;
 }
 
 - (NSOperation *)postCheckout:(NSDictionary *)checkoutJSON completion:(BUYDataCheckoutBlock)block
--- a/Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.h
@@ -30,8 +30,8 @@
 
 @property (strong, atomic) NSArray<NSOperation *> *operations;
 
-+ (instancetype)operationWithQueue:(NSOperationQueue *)queue operations:(NSArray<NSOperation *> *)operations;
-- (instancetype)initWithQueue:(NSOperationQueue *)queue operations:(NSArray<NSOperation *> *)operations;
++ (instancetype)aggregateOperationWithRequestQueue:(NSOperationQueue *)queue operations:(NSArray<NSOperation *> *)operations;
+- (instancetype)initWithRequestQueue:(NSOperationQueue *)queue operations:(NSArray<NSOperation *> *)operations;
 
 - (void)finishWithObject:(id)object;
 - (void)finishWithError:(NSError *)error;
--- a/Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.h
@@ -1,5 +1,5 @@
 //
-//  BUYCheckoutOperation.h
+//  BUYAggregateOperation.h
 //  Mobile Buy SDK
 //
 //  Created by Shopify.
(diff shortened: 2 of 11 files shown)


bgulanowski:
Was it really necessary to make this PR dependent on a previous PR? Would prefer that it be derived from the target branch (integration/sdk-2)

dbart01:
I rebased the branch on `integration/sdk-2`

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.m
```diff
@@ -152,6 +153,15 @@ - (NSOperation *)getCompletionStatusOfCheckoutWithToken:(NSString *)token start:
 	}];
 }
 
+- (NSOperation *)pollCompletionStatusAndGetCheckoutWithToken:(NSString *)token start:(BOOL)start completion:(BUYDataCheckoutBlock)block
```
bgulanowski: Since we really don't want to "leak" details about the underlying implementation in the method signature, perhaps `-getCompletedCheckoutWithToken:…` would be nicer.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.h
```diff
@@ -0,0 +1,39 @@
...
+//  THE SOFTWARE.
+//
+
+#import "BUYOperation.h"
+
+@interface BUYAggregateOperation : BUYOperation
+
+@property (strong, atomic) NSArray<NSOperation *> *operations;
+
++ (instancetype)operationWithQueue:(NSOperationQueue *)queue operations:(NSArray<NSOperation *> *)operations;
```
bgulanowski: I think we need to prefix "aggregate" in this method name. And maybe we should find a way to be clear that the `queue` is where request operations run, so `RequestQueue`?

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.m
```diff
@@ -0,0 +1,127 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+#import "BUYAggregateOperation.h"
+
+@interface BUYAggregateOperation ()
+
+@property (strong, nonatomic) NSOperationQueue *queue;
```
bgulanowski: This should not be strong.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
```diff
@@ -29,15 +29,15 @@
 #import "BUYClient+Internal.h"
 #import "BUYPaymentToken.h"
 #import "BUYRequestOperation.h"
+#import "BUYStatusOperation.h"
 
 @interface BUYCheckoutOperation ()
 
+@property (strong, nonatomic, readonly) BUYClient *client;
```
bgulanowski: This should probably not be a strong reference. Client is used only for communication. This object does not own it.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.m
dbart01: Nice catch.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
dbart01: Nope, you're right.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
```diff
@@ -144,25 +110,11 @@ - (NSOperation *)createBeginOperation
...
-	
-	return operation;
-}
-
-- (NSOperation *)createGetOperation
+- (NSOperation *)createStatusOperation
 {
-	return [self.client getCheckoutWithToken:self.checkoutToken start:NO completion:^(BUYCheckout *checkout, NSError *error) {
+	return [self.client pollCompletionStatusAndGetCheckoutWithToken:self.checkoutToken start:NO completion:^(BUYCheckout *checkout, NSError *error) {
 		if (checkout) {
```
bgulanowski: Instead of putting `cancelled` checks inside the finish methods, we could instead put them here. Would that simplify the code?

dbart01: I personally think its easier to read when we check for `cancelled` state in `finishWith...` methods. Its also less error prone since there's no chance that someone can forget to check for `cancelled` before calling `finishWith...`.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.m
```diff
@@ -0,0 +1,127 @@
...
+{
+	[super cancelExecution];
+	[self cancelAllOperations];
+}
+
+#pragma mark - Dependencies -
+
+- (void)linkDependencies
+{
+	if (self.operations.count > 1) {
```
bgulanowski: I think it would be cleaner to create a subarray from 1 to length-1, and then use a local variable to store the previous object in the array, and initialize that variable with the first object in the array. Then you won't need this conditional, because the subarray will simply have 0 elements if there is only one operation. 
`for (NSOperation *next in [operations subarrayWithRange:NSMakeRange(1, array.count - 1)];`

bgulanowski: Also, see `-[NSArray buy_tail]`.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
dbart01: On second thought, even though the operation doesn't own the client it wouldn't be a bad idea for the operation to keep a client alive if its in the process of being deallocated. If maintain only a weak reference to the client, it would then be possible that some internal operations are created and other are not (if the client is deallocated). This case would require explicit check and local strong references to the client making the code more complex. Since the strong reference doesn't (or shouldn't) unnecessarily keep the client alive in normal use cases, I would be in favour of keeping the strong reference.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.m
dbart01: What if there are `0` operations? It will raise an `NSRangeException` if the range is outside the bounds, which means that we'll need to check if the range is within bounds anyway.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
bgulanowski: If the client goes away, this operation should consider itself cancelled. If the client goes away, we can be certain that there are no listeners who care about the results of the work done by this class or its inherent operations, so it should stop working and discard its results.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.m
bgulanowski: OK. We will wait and see what the community has to say and if they ask for it or something similar.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYAggregateOperation.m
bgulanowski: `-buy_tail` checks for that. Create the tail first, and if it's empty, just bail out.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYCheckoutOperation.m
bgulanowski: Since this works without problems in practise, and has no functional risks that we can see, we can leave it for now.

Mobile Buy SDK/Mobile Buy SDK/Operations/BUYGroupOperation.m
```diff
@@ -0,0 +1,127 @@
...
+{
+	[super cancelExecution];
+	[self cancelAllOperations];
+}
+
+#pragma mark - Dependencies -
+
+- (void)linkDependencies
+{
+	if (self.operations.count > 1) {
```
bgulanowski:
```objective-c
NSArray *tail = self.operations.tail;
if (tail.count > 0) {
    NSOperation *prev = self.operations.firstObject; // we know this exists if tail has objects in it
    for (NSOperation *op in tail) {
        [op addDependency:prev];
        prev = op;
    }
}
```

bgulanowski: Or:
```objective-c
NSArray *tail = self.operations.tail;
NSOperation *prev = nil;
for (NSOperation *op in tail) {
    [op addDependency:prev];
    prev = op ?: self.operations.firstObject;
}
```

dbart01:
👍 ?

bgulanowski:
My PR seems to have caused conflicts in your PR.

bgulanowski:
:shipit:

bgulanowski:
Wait: don't merge until after #219 is merged

bgulanowski:
Ignore previous message - I was seeing outdated git diffs. (There is a duplication of the Xcode project changes made by Xcode, but that's fine.) :shipit: 🚀

PR #221, opened 2016-06-08
Update all public headers to use framework-style imports

bgulanowski:
# What

This converts use of project header import style (quotation marks) to use library/framework style (angle brackets).

# Why

All imports used by clients of a framework should use only the framework style, and likewise, any header imported should use that style. Project-style imports are meant for private headers and for implementation files.

The practical consequences are pretty minor. It ensures that if modules are disabled (and some users of the framework may have to disable modules in order to work with other source code that doesn't support it), then the framework will still work correctly.

@dbart01 @davidmuzi


dbart01:
🚢

PR #222, opened 2016-06-08
Automatic changes to project by Xcode.

bgulanowski:
Xcode apparently found some junk in the project that it should have cleaned up when the static framework target was removed.

@dbart01


bgulanowski:
Made redundant by https://github.com/Shopify/mobile-buy-sdk-ios/pull/225

PR #223, opened 2016-06-08
Update to new API for checkout creation

bgulanowski:
Models require a model manager when initialized.

@dbart01


dbart01:
🚢 :shipit:

PR #224, opened 2016-06-08
Add nullability annotations

dbart01:
### What this does
Adds nullability annotations for improved Swift-interoperability.

@bgulanowski @davidmuzi


bgulanowski:
There are still some incorrect import changes.

bgulanowski:
OK, this looks great. Snuck a few more bits and pieces in there, I see, but all good things. Thanks for fixing the warnings about relationships. We'll just have to keep an eye out for any new relationships to make sure they are well-formed. But it's not really critical for this kind of project. 👍

PR #227, opened 2016-06-08
Improve support in private methods

dbart01:
### What this does
- adds status to `BUYStatusOperation` completion for internal use
- moves `BUYClient` types into separate header

@bgulanowski @gabrieloc


bgulanowski:
LGTM

PR #228, opened 2016-06-08
Remove left over polling in BUYApplePayAuthorizationDelegate.

dbart01:
### What this does
Removes polling in `BUYApplePayAuthorizationDelegate`, a change that was somehow lost somewhere at some point during a merge...

@bgulanowski @gabrieloc

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
@@ -38,7 +38,7 @@
 
 const NSTimeInterval PollDelay = 0.5;
 
-typedef void (^AddressUpdateCompletion)(PKPaymentAuthorizationStatus, NSArray *shippingMethods, NSArray *summaryItems);
+typedef void (^AddressUpdateCompletion)(PKPaymentAuthorizationStatus, NSArray<PKShippingMethod *> * _Nonnull, NSArray<PKPaymentSummaryItem *> * _Nonnull);
 
 @interface BUYApplePayAuthorizationDelegate ()
 
@@ -132,13 +132,13 @@ typedef void (^AddressUpdateCompletion)(PKPaymentAuthorizationStatus, NSArray *s
 	[controller dismissViewControllerAnimated:YES completion:nil];
 }
 
--(void)paymentAuthorizationViewController:(PKPaymentAuthorizationViewController *)controller didSelectShippingAddress:(ABRecordRef)address completion:(void (^)(PKPaymentAuthorizationStatus, NSArray<PKShippingMethod *> * _Nonnull, NSArray<PKPaymentSummaryItem *> * _Nonnull))completion
+-(void)paymentAuthorizationViewController:(PKPaymentAuthorizationViewController *)controller didSelectShippingAddress:(ABRecordRef)address completion:(AddressUpdateCompletion)completion
 {
 	self.checkout.shippingAddress = [self buyAddressWithABRecord:address];
 	[self updateCheckoutWithAddressCompletion:completion];
 }
 
--(void)paymentAuthorizationViewController:(PKPaymentAuthorizationViewController *)controller didSelectShippingContact:(PKContact *)contact completion:(void (^)(PKPaymentAuthorizationStatus, NSArray<PKShippingMethod *> * _Nonnull, NSArray<PKPaymentSummaryItem *> * _Nonnull))completion
+-(void)paymentAuthorizationViewController:(PKPaymentAuthorizationViewController *)controller didSelectShippingContact:(PKContact *)contact completion:(AddressUpdateCompletion)completion
 {
 	self.checkout.shippingAddress = [self buyAddressWithContact:contact];
 	[self updateCheckoutWithAddressCompletion:completion];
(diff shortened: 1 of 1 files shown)


Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -177,9 +174,39 @@ - (void)updateCheckoutWithAddressCompletion:(void (^)(PKPaymentAuthorizationStat
 		[self.client updateCheckout:self.checkout completion:^(BUYCheckout *checkout, NSError *error) {
 			if (checkout && error == nil) {
 				self.checkout = checkout;
-				[self getShippingRates:self.checkout completion:completion];
-			}
-			else {
+				
+				if ([self.checkout requiresShipping] == NO) {
```
bgulanowski: I would be a lot happier with this if the completion block was a property of this object, so we didn't need such painful indentation. The logic in here is simply too hard to follow with all the nested conditionals.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -177,9 +174,39 @@ - (void)updateCheckoutWithAddressCompletion:(void (^)(PKPaymentAuthorizationStat
...
+					self.shippingRates = @[];
+					[self.client getShippingRatesForCheckoutWithToken:self.checkout.token completion:^(NSArray *shippingRates, BUYStatus status, NSError *error) {
+						
+						self.shippingRates = shippingRates;
+						if (shippingRates) {
+							
+							NSArray *shippingMethods = [BUYShippingRate buy_convertShippingRatesToShippingMethods:shippingRates];
+							if (shippingMethods.count > 0) {
+								
+								[self selectShippingMethod:shippingMethods[0] completion:^(BUYCheckout *checkout, NSError *error) {
```
bgulanowski: We should not need to pass completion blocks defined by self into methods owned by self. The `-selectShippingMethod:` method should be able to create this block, and will be able to if the caller's original completion block was available as an ivar or property.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -177,9 +174,39 @@ - (void)updateCheckoutWithAddressCompletion:(void (^)(PKPaymentAuthorizationStat
...
+				
+				if ([self.checkout requiresShipping] == NO) {
+					completion(PKPaymentAuthorizationStatusSuccess, nil, [self.checkout buy_summaryItemsWithShopName:self.shopName]);
+				} else {
+					
+					self.shippingRates = @[];
+					[self.client getShippingRatesForCheckoutWithToken:self.checkout.token completion:^(NSArray *shippingRates, BUYStatus status, NSError *error) {
+						
+						self.shippingRates = shippingRates;
+						if (shippingRates) {
```
bgulanowski: We don't need this conditional check—the `-buy_convertShippingRatesToShippingMethods:` will return an empty array if we pass in nil.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -177,9 +174,39 @@ - (void)updateCheckoutWithAddressCompletion:(void (^)(PKPaymentAuthorizationStat
...
+									completion(error ? PKPaymentAuthorizationStatusFailure : PKPaymentAuthorizationStatusSuccess, shippingMethods, [self.checkout buy_summaryItemsWithShopName:self.shopName]);
+								}];
+								
+							} else {
+								self.lastError = [NSError errorWithDomain:BUYShopifyError code:BUYShopifyError_NoShippingMethodsToAddress userInfo:nil];
+								completion(PKPaymentAuthorizationStatusInvalidShippingPostalAddress, nil, [self.checkout buy_summaryItemsWithShopName:self.shopName]);
+							}
+							
+						} else {
+							completion(PKPaymentAuthorizationStatusInvalidShippingPostalAddress, nil, [self.checkout buy_summaryItemsWithShopName:self.shopName]);
```
bgulanowski: Some code duplication in the error cases. Would it be worth trying to write an operation group for this?

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -177,9 +174,39 @@ - (void)updateCheckoutWithAddressCompletion:(void (^)(PKPaymentAuthorizationStat
...
+							if (shippingMethods.count > 0) {
+								
+								[self selectShippingMethod:shippingMethods[0] completion:^(BUYCheckout *checkout, NSError *error) {
+									if (checkout && error == nil) {
+										self.checkout = checkout;
+									}
+									completion(error ? PKPaymentAuthorizationStatusFailure : PKPaymentAuthorizationStatusSuccess, shippingMethods, [self.checkout buy_summaryItemsWithShopName:self.shopName]);
+								}];
+								
+							} else {
```
bgulanowski: It looks like this `else` block and the `else` in the enclosing block are both in the same scope, so possibly these could be collapsed.

Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
bgulanowski: We can collapse the first two conditionals here:
```objective-c
if (!error && [self.checkout requiresShipping]) {
  // normal case
}
else {
  self.lastError = error;
}
```

bgulanowski:
@dbart01 I've done some refactoring. Please have a look.

PR #233, opened 2016-06-16
Two model-level fixes

bgulanowski:
# Purpose

Fix two issues with the model.

1. incorrect delete rules on the Order - LineItem relationship.
2. incorrect property synthesis for a dynamic property defined in Core Data model. (This only affects builds that use the core data build option)

@dbart01

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
@@ -277,7 +277,7 @@
 		[operation cancel];
 	}];
 	
-	[self waitForExpectationsWithTimeout:4.0 handler:^(NSError *error) {}];
+	[self waitForExpectationsWithTimeout:5.0 handler:^(NSError *error) {}];
 }
 
 - (void)testCancellationWithoutQueue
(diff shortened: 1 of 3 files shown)


Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
```diff
@@ -277,7 +277,7 @@ - (void)testCancellationDuringExecution
 		[operation cancel];
 	}];
 	
-	[self waitForExpectationsWithTimeout:4.0 handler:^(NSError *error) {}];
+	[self waitForExpectationsWithTimeout:5.0 handler:^(NSError *error) {}];
```
dbart01: Just out of curiosity, is this intentional and / or does it fix the BuildKit crash we've been experiencing?

dbart01:
LGTM 🚢

Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
bgulanowski: I took a guess that this was causing the test crash, and it seemed to help.

PR #236, opened 2016-06-20
Mark buy_emailFromRecord as deprecated

abecevello:
The `buy_emailFromRecord:` method in BUYApplePayAdditions.h uses `ABRecordRef` which is deprecated by Apple as of iOS 9.0. Similar methods in the header are marked as deprecated but it looks like this one was missed. This PR just adds the proper deprecation markers so Xcode doesn't complain when using the framework on a iOS 9.0+ project.

Looking at what the method was doing, I don't believe a `buy_` replacement is necessary. Getting the email address from a `PKContact` is a single property access, as opposed to `ABRecordRef` which is lots of fun to get stuff out of.

Please review @bgulanowski


bgulanowski:
Looks good to me! 🚢 it.

PR #243, opened 2016-06-23
Improve consistency check when starting checkout

bgulanowski:
# What

This allows checkout to continue if already started, if the checkout being passed in matches the one currently in progress.

# Why

This allows an app to continue an existing web checkout that the user interrupted. On iOS 8, when starting a checkout, we open Safari. If the user returns from Safari without finishing their checkout, perhaps to modify their cart, we should allow checkout to continue.

An alternative approach would be to put more onus on the app to make sure that the payment provider state is reset, but this is simple and should be safe.

@davidmuzi @gabrieloc


gabrieloc:
🎩

PR #246, opened 2016-06-27
Replace generated accessors with method redirection

bgulanowski:
# Why

## Background

When implementing a caching solution, you have to decide whether model objects used in memory should be the same objects used by the caching system, or to use different objects. In the case of Core Data, it means whether to use objects inherited from `NSManagedObject`. Managed objects are designed to be used in memory, and there are many benefits to doing so.

We chose to make it straightforward to add support for Core Data caching, since we needed it ourselves. But we didn't want to force other users of the iOS Buy SDK to do so.

As a solution, we provide a build setting which alters the inheritance for model objects. If Core Data support is enabled, model classes inherit from `BUYManagedObject` (which inherits from `NSManagedObject`). Otherwise, they inherit from `BUYObject` (which inherits from `NSObject`). The mechanism for this is to define a macro, `BUYCachedObject`, which resolves to either `BUYManagedObject` or `BUYObject`, depending on whether Core Data support is enabled.

But this leads to a complication when you want to provide custom implementations of property accessor methods. `NSManagedObject`, and classes which inherit from it, will only generate dynamic accessors if none exist. `NSManagedObject` explicitly opts out of compiler support for synthesized accessors, and this behaviour is inherited. If you write a custom accessor, you must follow a specific pattern, sending access notifications with `-willAccessValueForKey:` and `-didAccessValueForKey:` (for getters) or change notifications (for setters), and using the "primitive" KVC accessor methods, `-primitiveValueForKey:` and `-setPrimitiveValue:forKey:`.

## Solutions

One solution for this is to implement all custom accessors in the Core Data style, and then somehow make that work in classes descended from `BUYObject`. While empty implementations of the access notification methods are simple enough, implementing the primitive KVC methods is not. If you, tried, for instance, to implement them as cover methods for the standard KVC accessors, you would cause a recursive loop. The alternative is to hack in the runtime:
- perform a dynamic lookup on the superclass's methods and perform a directly message send (discouraged by the documentation)
- perform a dynamic lookup of the instance variable used to back the property, and write directly to it (just thought of that, kind of interesting)

# What

A third solution is possible if we continue implement all custom accessors in the standard `NSObject` style, by wrapping `super`'s implementation, which exists when Core Data support is not enabled. When Core Data is enabled, use forwarding to make it appear that `super`'s implementation exists, even when it doesn't.

# How

This requires using the runtime in a more acceptable way, by detecting accessor message sends, and diverting them to Core Data primitive KVC methods.

In this PR, I have implemented this approach, as a proof of concept. It adds 75 lines of well-factored code to intercept message sends to super, which would otherwise fail and cause an exception when the implementation is not found.

We use the forwarding machinery supported by `NSObject`. We only have to use a runtime function in once case, to identify whether a message is trying to get or set a property. We don't change any classes.

As with all forwarding work, we have to implement three methods: `-respondsToSelector:`, `-methodSignatureForSelector:`, and `-forwardInvocation:`. (For more information on message forwarding in Objective-C, see the [Objective-C Runtime Programming Guide](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtForwarding.html).)

We also need some fairly simple utilities: a function to convert selector names into property names (specifically for converting `-setKey:` into `key`), and two methods to test whether a selector represents a property accessor.

Finally, we need to implement methods which implement the "primitive" KVC accessor pattern.

In the more normal forwarding case of sending an invocation to a different object (a technique that might be called "dynamic cover methods"). But here, the recipient is `self`, and we want to use different methods to read and write the values.

To make this work, we use the method signatures for the "primitive" KVC accessors, and we use those methods directly in `-forwardInvocation:`, instead of literally forwarding.

## Fourth way

I explored another alternative which avoids using the runtime in favour of some fairly simple macros, but they make the code ugly. Briefly, the macros would look like:

```objective-c
#ifdef CORE_DATA_PERSISTENCE
#define WillAccess(key) ([self willAccessValueForKey:key])
#define DidAccess(key) ([self didAccessValueForKey:key])
#define WillChange(key) ([self willChangeValueForKey:key])
#define DidChange(key) ([self didChangeValueForKey:key])
#define GetSuperProperty(key) ([self primitiveValueForKey:key])
#define SetSuperProperty(key, unused, value) ([self setPrimitiveValue:value forKey:key])
#else
// deliberately empty definitions
#define WillAccess(key)
#define DidAccess(key)
#define WillChange(key)
#define DidChange(key)
#define GetSuperProperty(key) ([super key])
// Hand-waving magic for converting "key" to "setKey:"
#define SetSuperProperty(key, setter, value) ([super setter value])
#endif

#define GetPrimitiveProperty(key) do {\
  WillAccess(key);\
  id value = GetSuperProperty(key);\
  DidAccess(key);\
  return value;\
} while(0)

#define SetPrimitiveProperty(key, setter, value) do {\
WillChange(key);\
SetSuperProperty(key, setter, value);\
DidChange(key);\
} while(0)

// An example method
- (NSString *)countryCode
{
  return [GetPrimitiveProperty(countryCode) uppercaseString];
}

- (void)setCountryCode:(NSString *)countryCode
{
  SetSuperProperty(countryCode, setCountryCode:, [countryCode lowercaseString]);
}
```

The `SetSuperProperty()` macro is the ickiest part, but using macros is in general pretty icky. Adding support for dynamically converting `countryCode` to `setCountryCode` is possible, but is almost as complicated as the message forwarding solution, and makes using macros at all basically pointless, because you have to perform the message send "manually", in defiance of the documentation.

I submit this as an interesting exploration of the Objective-C runtime. I'm not concerned whether we use it or not.

cc: @g-Off @ChristoDeluxe


Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
```diff
@@ -91,6 +92,83 @@ - (NSDictionary *)jsonDictionaryForCheckout
...
+	}
+	return setter ? [[name substringWithRange:NSMakeRange(3, name.length - 4)] lowercaseString] : name;
+}
+
+- (BOOL)respondsToSelector:(SEL)selector
+{
+	return [super respondsToSelector:selector] || [self isPropertyAccessor:selector];
+}
+
+static NSMutableDictionary *signatureCache;
```
dbart01: I'm not really a fan of file variables. Would personally prefer to see this in a class method instead (but just my opinion):
```objc
+ (NSMutableDictionary *)signatureCache
{
    static NSMutableDictionary *cache = nil;
    static dispatch_once_t onceToken;
	dispatch_once(&onceToken, ^{
        cache = [NSMutableDictionary new];
	});
    return cache;
}
```

dbart01:
Very nice 👍  Love the removal of all those accessors!

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
```diff
@@ -91,6 +92,83 @@ - (NSDictionary *)jsonDictionaryForCheckout
...
+#pragma mark - Dynamic Property Accessor Resolution
+
+NS_INLINE NSString *KeyForSelector(SEL selector, BOOL *isSetter)
+{
+	NSString *name = NSStringFromSelector(selector);
+	BOOL setter = [name hasPrefix:@"set"];
+	if (isSetter) {
+		*isSetter = setter;
+	}
+	return setter ? [[name substringWithRange:NSMakeRange(3, name.length - 4)] lowercaseString] : name;
```
g-Off: can simplify this with `substringFromIndex:`

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
g-Off: Is the signature cache actually necessary? The runtime does do some of its own caching but I'm not sure exactly when that happens and under what circumstances

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
```diff
@@ -91,6 +92,83 @@ - (NSDictionary *)jsonDictionaryForCheckout
...
+	dispatch_once(&onceToken, ^{
+		signatureCache = [NSMutableDictionary dictionary];
+	});
+	
+	NSString *selectorName = NSStringFromSelector(selector);
+	NSMethodSignature *signature = signatureCache[selectorName];
+	
+	if (!signature) {
+		SEL substituteSelector = [selectorName hasPrefix:@"set"] ? @selector(setPrimitiveValue:forKey:) : @selector(primitiveValueForKey:);
+		signatureCache[selectorName] = signature = [super methodSignatureForSelector:substituteSelector];
```
g-Off: for readability this would be better as two separate statements

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
```diff
@@ -91,6 +92,83 @@ - (NSDictionary *)jsonDictionaryForCheckout
...
+}
+
+- (BOOL)isPropertyAccessor:(SEL)selector
+{
+	return (signatureCache[NSStringFromSelector(selector)] || [self hasPropertyForKey:KeyForSelector(selector, NULL)]);
+}
+
+- (BOOL)hasPropertyForKey:(NSString *)key
+{
+	return class_getProperty([self class], [key UTF8String]) != NULL;
```
g-Off: could use property style syntax for `key.UTF8String`

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
bgulanowski: Except there's that colon at the end 😞

Mobile Buy SDK/Mobile Buy SDK/Models/BUYManagedObject.m
bgulanowski: I did not profile it, so it is premature optimization. The speed improvement from the cache is probably marginal for generating the signatures. It is also used to return quickly from `-isPropertyAccessor:`.

davidmuzi:
Will this PR be updated, or should it be closed?

bgulanowski:
No one has objected to the design, so I will rebase and merge it.

PR #249, opened 2016-07-08
Remove public product url

davidmuzi:
### What

This removes the public URL from the product entity, as it is not sent by Shopify

@bgulanowski


bgulanowski:
🎩

PR #251, opened 2016-07-12
Fix customer app

davidmuzi:
The customer sample app is written in Swift, and has an extension on BUYOrder to type the array of line items (which is exposed as an NSOrderedSet in the SDK.  To avoid a naming conflict, this change modifies the method name in the sample app.

@bgulanowski


Mobile Buy SDK Sample Apps/Sample App Customers/Sample App Customers/OrdersViewController.swift
```diff
@@ -108,7 +108,7 @@ extension OrdersViewController: UITableViewDataSource {
 //
 extension BUYOrder {
     
-    var lineItemsArray: [BUYLineItem] {
+    var arrayOfLineItems: [BUYLineItem] {
```
bgulanowski: You can remove this method outright. `-lineItemsArray` exists in the SDK.

bgulanowski:
🎩  - no need to rebuild. The failing test is flaky. We will have to rewrite it.

Issue #253, opened 2016-07-14
Remove password_confirmation field from request payload


mtaher:
The customer APIs were changed to not required the `password_confirmation` field.

The SDK should be updated accordingly.

This affects the following endpoints,
- Customer creation
- Customer activation
- Customer password reset

mtaher:
cc @gabrieloc @davidmuzi @bgulanowski

PR #259, opened 2016-07-15
Remove password confirmation

davidmuzi:
### What this does

- Removes the password confirmation key from customer APIs, as this is a client concern.
- Removes associated keys from unit tests
- Updates documentation

Fix: #253 

@bgulanowski @gabrieloc


bgulanowski:
Looks good. 🍏

PR #260, opened 2016-07-15
Use NSNumber for product id in Client APIs

davidmuzi:
The type for ID in the product model is NSNumber.  We should use the same type in the APIs.

@bgulanowski


bgulanowski:
👍

PR #267, opened 2016-07-20
Introduce BUYCustomerToken and update APIs

bgulanowski:
# Purpose

The customer authentication (token creation) endpoint provided by Shopify returns an access token (`customer_access_token`) object. This includes three attributes: `customer_id`, `expires_at` and `access_token`. Up until now, the buy client has been keeping a reference to the token string, using the customer ID to fetch the customer, and throwing away the expiry date.

On Android, in contrast, we keep a reference to the entire access token.

As a consequence, APIs that require a customer to be authenticated can use the `customerID` from the access token, instead of a needing a parameter.

# Approach

This PR updates the behaviour to match Android, keeping a reference to the entire access token.

We have renamed a private class to `BUYCustomerToken` and made it public. Callbacks now return this object, instead of the token string.

Simultaneously, we are updating customer APIs on both Android and iOS to remove the `customerID` parameter for those APIs that require authentication: getting and updating the customer, renewing the token, and all address access.

Fixes https://github.com/Shopify/mobile-buy-sdk-ios/issues/268

# Steps

- [ ] Unit tests
- [ ] Code Review
- [ ] Test

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
@@ -49,6 +49,7 @@
 - (NSURL *)urlForCustomers;
 - (NSURL *)urlForCustomersToken;
 - (NSURL *)urlForCustomersPasswordRecovery;
+- (NSURL *)urlForLoggedInCustomer;
 
 - (NSURL *)urlForCustomersOrdersWithID:(NSString *)identifier;
 - (NSURL *)urlForCustomersOrdersWithID:(NSString *)identifier orderID:(NSNumber *)orderID;
--- a/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
@@ -50,13 +50,14 @@
 - (NSURL *)urlForCustomersToken;
 - (NSURL *)urlForCustomersPasswordRecovery;
 - (NSURL *)urlForLoggedInCustomer;
+- (NSURL *)urlForLoggedInCustomerToken;
+- (NSURL *)urlForCustomersTokenRenewal;
 
-- (NSURL *)urlForCustomersOrdersWithID:(NSString *)identifier;
-- (NSURL *)urlForCustomersOrdersWithID:(NSString *)identifier orderID:(NSNumber *)orderID;
+- (NSURL *)urlForCustomersOrders;
+- (NSURL *)urlForCustomersOrdersWithOrderID:(NSNumber *)orderID;
+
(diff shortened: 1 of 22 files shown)


bgulanowski:
@dbart01 @gabrieloc @davidmuzi

dbart01:
Do we want to rename method names that end with `Callback` to `WithCallback`? For example, in the case of:
```objc
- (void)renewCustomerTokenCallback:(BUYDataTokenBlock)block;
```
It sounds like we're renewing the callback instead of the customer token. Would `renewCustomerTokenWithCallback` sound better?

bgulanowski:
For callbacks aka completion blocks, no, I prefer not to have the "With".

Mobile Buy SDK/Mobile Buy SDK/Models/BUYCustomerToken.m
```diff
@@ -71,4 +81,9 @@ - (NSDictionary *)JSONDictionary
 			 };
 }
 
+- (NSString *)JSONString
```
davidmuzi: need this?

davidmuzi:
Unit tests need updating

davidmuzi:
Customer Sample App needs updating too

bgulanowski:
The only thing failing now is the same flaky test that has been a problem for a while now.

Other issues have been addressed.

gabrieloc:
Unrelated, but it would appear `BUYOrder` is no longer getting `statusURL` assigned. It looks like the response changed to come back with `orderStatusURL` instead.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
```diff
@@ -49,16 +49,18 @@
...
+- (NSURL *)urlForLoggedInCustomer;
+- (NSURL *)urlForLoggedInCustomerToken;
+- (NSURL *)urlForCustomersTokenRenewal;
+
+- (NSURL *)urlForCustomersOrders;
+- (NSURL *)urlForCustomersOrdersWithOrderID:(NSNumber *)orderID;
 
-- (NSURL *)urlForCustomersOrdersWithID:(NSString *)identifier;
-- (NSURL *)urlForCustomersOrdersWithID:(NSString *)identifier orderID:(NSNumber *)orderID;
 - (NSURL *)urlForCustomersWithID:(NSString *)identifier;
```
davidmuzi: Why aren't these changed?

bgulanowski:
Was determined that the `orderStatusURL` problem was localized to a few old orders on a test shop. Not impacted by this PR in any way.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Routing.h
bgulanowski: Just removed 😄

Issue #272, opened 2016-07-22
Full support for customer update


bgulanowski:
Should be able to change the addresses and default address.

PR #273, opened 2016-07-22
Add customer update endpoint

davidmuzi:
### What is this

Fix: #272 

Adds support for the update customer endpoint.  Removes pointless customer update unit test.

@bgulanowski

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYClient+CustomerTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYClient+CustomerTests.m
@@ -197,6 +197,23 @@
 	[self waitForExpectationsWithTimeout:10.0 handler:^(NSError *error) {}];
 }
 
+- (void)testCustomerUpdate
+{
+	[OHHTTPStubs stubUsingResponseWithKey:@"testCustomerLogin" useMocks:[self shouldUseMocks]];
+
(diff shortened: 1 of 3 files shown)


Mobile Buy SDK/Mobile Buy SDK Tests/BUYClient+CustomerTests.m
```diff
@@ -197,29 +197,6 @@ - (void)testCustomerOrderWithID
 	[self waitForExpectationsWithTimeout:10.0 handler:^(NSError *error) {}];
 }
 
-#pragma mark - Update -
-
-- (void)testCustomerUpdate
```
davidmuzi: Is there a point in mocking an `update` method?

bgulanowski: To make sure that the callback is invoked, and with the right kinds of objects?

bgulanowski: When running this against Shopify, shouldn't we check, for example, that changing the default address does the right thing? If we had two, we could alternate, although then the test wouldn't be stateless (unless we changed it and then changed it back).

bgulanowski:
Lookin' good! 🎩

PR #278, opened 2016-07-26
Add "Address.isDefault" boolean <-> "default"

### LINKED ISSUE #277: Addressses have no "default" property
Addresses from Shopify have a "default" property. Add "isDefault" to Address entity.

bgulanowski:
# What

Add `isDefault` attribute to `Address` entity.

We will add additional conveniences for making sure that setting the default address on a customer behaves nicely. For now, change the value directly on the correct address, and ensure any other addresses are no longer marked as default.

Fixes #277 

@gabrieloc @krisorr


krisorr:
FWIW, looks good to me

PR #280, opened 2016-07-28
Args are in wrong order.

bgulanowski:
This puts arguments in the correct order when invoking the completion block.

This causes a compiler error when built for device:

```
mobile-buy-sdk-ios/Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m:140:48: error: implicit conversion loses integer precision: 'const NSUInteger' (aka 'const unsigned int') to 'BOOL' (aka 'signed char') [-Werror,-Wconversion]
                block(tags, [self hasReachedEndOfPage:tags], page, error);
                ~~~~~                                        ^~~~
```
But this issue is not caught when building for simulator.

@gabrieloc @davidmuzi


bgulanowski:
ping @gabrieloc

gabrieloc:
Surprised this wasn't caught earlier. 👍

PR #283, opened 2016-07-31
Removed Swift sample app link in README, and replaced broken links

dengjeffrey:
Removed Swift sample app link in README, and replaced broken links.

The new sample app links to their respective folders instead of README.md which the old links linked too, since their respective folders display the README.md anyways

@bgulanowski @gabrieloc


gabrieloc:
👍 Fixes issue

PR #287, opened 2016-08-05
change completeCheckout: return value from BUYOperation to NSOperation

horsejockey:
completeCheckout: was returning a private class BUYOperation.  When using the Buy framework in Swift the completeCheckout: method was not available because of the private class return type.  I changed the return value for the method to NSOperation which lets the method be available.  Alternatively you could make BUYOperation public.


davidmuzi:
LGTM

PR #292, opened 2016-08-09
Get Collections by ID

gabrieloc:
### What this does
Allows collections to be retrieved by an array of IDs

@bgulanowski


bgulanowski:
Needs a test.

bgulanowski:
@krisorr do you guys have this on Android? Seems a no-brainer. Not sure how we missed it.

krisorr:
@bgulanowski - yes, we added it after the 2.0 release, it is on develop

davidmuzi:
How does this work when the number of collections specified exceeds the page size?

gabrieloc:
@bgulanowski Added tests
@davidmuzi There's API for getting paginated and unpaginated collections. Added support for both.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m
```diff
@@ -173,6 +173,27 @@ - (NSOperation *)getCollectionsPage:(NSUInteger)page completion:(BUYDataCollecti
 	}];
 }
 
+- (NSOperation *)getCollectionsByIds:(NSArray<NSString *> *)collectionIds completion:(BUYDataCollectionsBlock)block
```
krisorr: do you need this one? it is redundant.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m
```diff
@@ -173,6 +173,27 @@ - (NSOperation *)getCollectionsPage:(NSUInteger)page completion:(BUYDataCollecti
...
+{
+	return [self getCollectionsByIds:collectionIds page:0 completion:block];
```
davidmuzi: also, page starts at `1`

gabrieloc: This is probably unclear. Asking for page 0 doesn't get page 1, it gets all pages. @mtaher Can you confirm?

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m
krisorr: We will need to update the Android SDK, it is actually ignoring the page and only doing getCollectionsByIds(List<> collectionIds)

gabrieloc: See comment below

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m
krisorr: I don't think so, I believe asking for page 0 == page 1

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m
gabrieloc: You're correct actually, I'll remove this.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Storefront.m
mtaher: > I don't think so, I believe asking for page 0 == page 1

Yup 👍 

@gabrieloc, are you seeing a different behaviour because you're not specifying a `limit`?

gabrieloc:
@davidmuzi I've made page number required, to make it more obvious that this api is paginated.

bgulanowski:
`-testCancellationDuringExecution` is the only test failure, which is sporadic.

👍

krisorr:
@gabrieloc @bgulanowski 
@sav007 
One issue that has come up.  This is now inconsistent with getProductsByIds.  I would expect that that API is paginated as well, but it does not accept a page number.

bgulanowski:
If we had been inconsistent the other way, maybe I wouldn't have cared as much. But we probably do want to paginate products, even though asking for more than 25 by id seems very unusual.

PR #298, opened 2016-08-18
Customer creation tests

gabrieloc:
## What this does
Adds a unit test for customer creation

@bgulanowski @krisorr

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYClient+CustomerTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYClient+CustomerTests.m
@@ -43,12 +43,6 @@
 
 #pragma mark - Lifecycle -
 
-- (void)setUp
-{
-	[super setUp];
-	[self loginDefaultCustomer];
-}
-
 - (void)tearDown
 {
 	[super tearDown];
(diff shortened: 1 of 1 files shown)


Mobile Buy SDK/Mobile Buy SDK Tests/BUYClient+CustomerTests.m
```diff
@@ -424,4 +458,12 @@ - (BUYAccountCredentials *)credentialsForFailure
 	return [BUYAccountCredentials credentialsWithItems:@[email, password]];
 }
 
+#pragma mark - Conveniences -
+
+- (NSString *)randomStringWithLength:(NSInteger)length
+{
+	NSString *UUID = [[NSUUID UUID] UUIDString];
```
bgulanowski: While every UUID is basically guaranteed to be unique, substrings are not. Does that affect your design?

gabrieloc: We're using the first 8 characters for emails (same as android), which should be reliable enough.

bgulanowski:
Didn't mean to neglect this, but assuming new tests are passing, let's merge.

Issue #304, opened 2016-08-23
Advanced Sample App fails to show Apple Pay button


davidmuzi:
Steps to reproduce:
1. setup shop credentials
2. set MERCHANT_ID in code, and in project
3. present BUYProductViewController on device with Apple Pay capability
4. Apple Pay button is not visible

PR #305, opened 2016-08-23
Fix Apple Pay button logic in PaymentViewController

davidmuzi:
Fix: #304 

### What

This fixes a bug in the sample app that prevents the Apple Pay button from appearing

### How

Removes a duplicate call to setup the button, which was using the wrong boolean value.

@bgulanowski


bgulanowski:
Tested. Button does appear for products that are in stock. 🚢  it.

PR #306, opened 2016-08-25
Xcode 8 Support

davidmuzi:
Add some required casting

@bgulanowski


gabrieloc:
We should do a find/replace for `BUYObject` -> `BUYCachedObject`

bgulanowski:
LGTM 🎩 

@gabrieloc it's not that simple.

First, `BUYObject` is a protocol and a class. It's possible that in more places we should be using `id<BUYObject>` or `NSObject<BUYObject> *`. Although most, if not all, references to models in the supporting model code use this already.

Second, the transient model objects all must inherit from `BUYObject` (the class), not `BUYManagedObject`. (The conditionally persistent models, on the other hand, inherit form `BUYCachedObject` because of their conditional persistence.)

Issue #315, opened 2016-08-31
Cannot debug Swift code in some situations


bgulanowski:
At Shopify, we use the Buy SDK from within another framework, in turn used by our app. When trying to debug swift, the debugger fails to read all the symbols:

```
@import Buy;
        ^
<module-includes>:1:9: note: in file included from <module-includes>:1:
#import "Headers/Buy.h"
        ^
/Users/Shared/Xcode/Derived/Merchant_App-axhqzbxtmaiuvbbaxcfsmpmxrbjj/Build/Products/Debug-core-data-iphonesimulator/Buy.framework/Headers/Buy.h:36:9: note: in file included from /Users/Shared/Xcode/Derived/Merchant_App-axhqzbxtmaiuvbbaxcfsmpmxrbjj/Build/Products/Debug-core-data-iphonesimulator/Buy.framework/Headers/Buy.h:36:
#import "BUYCollection.h"
        ^
/Users/Shared/Xcode/Derived/Merchant_App-axhqzbxtmaiuvbbaxcfsmpmxrbjj/Build/Products/Debug-core-data-iphonesimulator/Buy.framework/Headers/BUYCollection.h:28:9: note: in file included from /Users/Shared/Xcode/Derived/Merchant_App-axhqzbxtmaiuvbbaxcfsmpmxrbjj/Build/Products/Debug-core-data-iphonesimulator/Buy.framework/Headers/BUYCollection.h:28:
#import "BUYClient+Storefront.h"
        ^
/Users/Shared/Xcode/Derived/Merchant_App-axhqzbxtmaiuvbbaxcfsmpmxrbjj/Build/Products/Debug-core-data-iphonesimulator/Buy.framework/Headers/BUYClient+Storefront.h:27:9: error: 'Buy/BUYClient.h' file not found
#import <Buy/BUYClient.h>
        ^
<module-includes>:2:9: note: in file included from <module-includes>:2:
#import "Headers/AsyncImageView.h"
        ^
/Users/brent/Dev/github/Shopify/mobile-buy-app-ios/Core/Core/AsyncImageView.h:28:9: error: could not build module 'Buy'
@import Buy;
```

@gabrieloc

bgulanowski:
Fixed by #316

PR #316, opened 2016-08-31
Convert all #import declarations to use user/quoted style.

bgulanowski:
The correct way to declare dependencies in public headers of public frameworks is to use framework style: `<Framework/File.h>`. This is to ensure compatibility with projects not using modules, although, these days, Xcode probably works around that some other way.

However, either because of a bug in lldb, or because of community members voting with their feet, the standard now is to use the user-style `#import` directives (with quotation marks). Perhaps as a result of this, lldb does not work with framework-style import directives.

This converts to using user-style, to accommodate implicit expectations of lldb.

Fixes https://github.com/Shopify/mobile-buy-sdk-ios/issues/315

@gabrieloc @davidmuzi @dbart01

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYAccountCredentialsTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYAccountCredentialsTests.m
@@ -24,8 +24,8 @@
 //  THE SOFTWARE.
 //
 
-#import <XCTest/XCTest.h>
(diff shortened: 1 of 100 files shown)


bgulanowski:
Still failing on a test due to inexplicable crash (`-testCancellationDuringPolling`). All other tests pass.

gabrieloc:
Not much to review, 🎩

PR #317, opened 2016-08-31
Get collections by handles

gabrieloc:
## What this does
Adds support for getting multiple collections by multiple handles

@bgulanowski


bgulanowski:
Should have a test.

gabrieloc:
Looks like this API isn't available after all.

Issue #313, opened 2016-08-31
BUYShippingRate.shippingRateIdentifier is nil on iphone 5


mihai-moldovan:
I'm using shippingRateIdentifier in the UI code to keep track of the selected shipping rate until the user applies the changes. However shippingRateIdentifier is nil on iphone 5 (device and simulator) on all tested ios versions (8.4, 9.3) 
Also it appears that BUYShippingRate.identifier is a String instead of NSNumber:

```
(lldb) p rate.identifier
(NSNumber!) $R4 = (null) {
  Foundation.NSValue = shopify-US:%202-4%20business%20days%20-%20Under%20$4,000-15.00
}
```
I attached a sample project where you can reproduce the issue:  [BuyTest.zip](https://github.com/Shopify/mobile-buy-sdk-ios/files/446546/BuyTest.zip)

Please add your own domain, api key and product id

mihai-moldovan:
This also happens on iphone 4S. I'm using sdk version 2.0

bgulanowski:
This seems to imply that it works on iPhone 5S and better. We don't have any code which checks the OS version. I'm not aware of any code which relies on 64-bit values, although it's possible.

Keep in mind that Shopify objects have identifiers that are 64 bits. We have tried to make everything an NSNumber so that any 64-bit values coming down the wire are handled correctly, but if you have any insight into where this might not be the case, please let us know.

mihai-moldovan:
OK, I investigated more. So I don't think it's related to any platform or OS version, although it easier to reproduce on iphone 5 with os version 8.x. 

Anyway I think I found the problem: ShippingRate entity defines JSONPropertyKey for shippingRateIdentifier as being "id"
[In NSEntityDescription+BUYAdditions.m](https://github.com/Shopify/mobile-buy-sdk-ios/blob/master/Mobile%20Buy%20SDK/Mobile%20Buy%20SDK/Additions/NSEntityDescription%2BBUYAdditions.m#L111) 
```
// json key -> property name
- (NSDictionary *)buy_JSONKeyMapping
{
	return [[self buy_PropertyNameMapping] buy_reverseDictionary];
}
```
The first call will return a dictionary like this:
```
{
    "identifier": "id",
    "shippingRateIdentifier": "id"
}
```
which is then reversed using buy_reverseDictionary. This, in turn will yield an undefined result. 
Sometimes it is:
`{ "id": "shippingRateIdentifier" }`
other times it is:
[...]

PR #318, opened 2016-09-01
Remove `identifier` property from `ShippingRate` entity

bgulanowski:
`ShippingRate` was incorrectly given an `identifier` property when the model was converted for SDK version 2. `identifier` may have been intended to replace `shippingRateIdentifier`, but that would have been inappropriate in any case, as the `shippingRateIdentifier` is a string, not a number.

Fixes https://github.com/Shopify/mobile-buy-sdk-ios/issues/313

cd @davidmuzi @gabrieloc for review


bgulanowski:
Failing test is unrelated.

davidmuzi:
should point to develop

bgulanowski:
I think we should get into the habit of making hot fixes on master. Otherwise we have to ship everything on develop in order to release a fix like this.

davidmuzi:
hot fixes should be put on a release branch stemming from the previous
release, then it will go back to master

bgulanowski:
Will re-open on release/2.0.1

PR #319, opened 2016-09-01
Remove `identifier` property from `ShippingRate` entity.

bgulanowski:
`ShippingRate` was incorrectly given an `identifier` property when the model was converted for SDK version 2. `identifier` may have been intended to replace `shippingRateIdentifier`, but that would have been inappropriate in any case, as the `shippingRateIdentifier` is a string, not a number.

Fixes https://github.com/Shopify/mobile-buy-sdk-ios/issues/313

cd @davidmuzi @gabrieloc for review


davidmuzi:
LGTM

PR #321, opened 2016-09-01
Remove code from sample app that is not used

bgulanowski:
Left over from the updates for 2.0.

@davidmuzi @gabrieloc


gabrieloc:
I dont understand how that code isn't needed. Should it be exercised?

bgulanowski:
The product view controller configures those buttons itself, including setting their labels and actions. The wrapper wasn't really adding anything, and it was inconsistent, because the abstraction (of the buttons) was incomplete.

bgulanowski:
Could cherry pick this into `release/2.0.1`.

davidmuzi:
🚢

PR #322, opened 2016-09-01
fix product fetch signature in sample app

davidmuzi:
The method signature for fetching products has changed in 2.0, this PR updates the call in the Advanced Sample App

@bgulanowski


bgulanowski:
If you rebase, you should get the update to `-testCustomerInvalidEmailPassword`, which looks like it's failing here.

PR #327, opened 2016-09-03
Feature: fetch all pages

bgulanowski:
Convenience for getting everything. Observable. Cancellable.

Support for collections, products, and tags.

⚠️   Use with caution! ⚠️ 

@davidmuzi @gabrieloc

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYAllPagesOperationTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYAllPagesOperationTests.m
@@ -1,0 +1,46 @@
+//
+//  BUYAllPagesOperationTests.m
+//  Mobile Buy SDK
+//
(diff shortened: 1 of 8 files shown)


davidmuzi:
Do we need this? What about Android?

krisorr:
@davidmuzi no, I don't think we need it. I would be hesitant to add this to the sdk.

gabrieloc:
We need proper pagination support instead

bgulanowski:
How do we feel about making the operation classes public?

pwr7549:
We need to fetch all products and collections. If we collect these product_types, we will be able to create a list of product_types at the opening on Apple TV. We also could call the certain collections or products by a specified product_type.

bgulanowski:
@pwr7549 please add a github story, and we can label it as an enhancement request.

PR #328, opened 2016-09-03
Check whether many-to-many relationship allows encoding.

### LINKED ISSUE #326: JSON Caching Issue with  NSSet
I am having an issue caching the BUYProduct.  I am encoding the JSONDictionary object to Data and storing to disk then reading it back out again.  However, I think there is an issue writing the NSMutableSet properties like the BUYProductVariant options.  When I load these from the cache the property is nil.  NSOrderedMutableSet works fine.  This makes sense as arrays can be converted without any effort to JSON where Sets can't.  I am not an expert though so maybe I am missing something.

bgulanowski:
Fix issue where product variants did not encode their option values.

Fixes #326 

@davidmuzi @gabrieloc


bgulanowski:
This PR is invalid. The original code was correct. Instead, ProductVariant should encode its options as ids. Options will have to be encoded separately. Custom caching solutions will have to re-form the relationships by implementing logic to convert the ids back to objects.

PR #329, opened 2016-09-03
Stricter approach to many-to-many JSON encoding.

bgulanowski:
The previous solution is simple, but more lenient in the case of inconsistent models. We should have automated verification to ensure every relationship is defined to prevent circular encoding or missing values.

The simple fix (targeting a bugfix release 2.0.1) checks whether the inverse relationship allows inverse encoding, but it is both convoluted and too lenient.

We have an implicit rule in our model that any relationship should specify ownership via implication of delete rules: if a relationship deletes its destination, then it owns it. If it owns it, the inverse should not encode itself, because it would be encoding its owner. But if a relationship's ownership cannot be inferred, then we also will encode from either direction, which could lead to a circular (recursive) encoding, as the same relationship is traversed back and forth.

There should not be any relationships without ownership in our model. Every relationship should have an owner. But we need to verify, and assert this assumption.

This PR includes a deprecation.
This PR also includes a change to convert a method into a read-only property.

Replacement for https://github.com/Shopify/mobile-buy-sdk-ios/pull/328

@davidmuzi @gabrieloc

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Additions/NSEntityDescription+BUYAdditions.h
+++ b/Mobile Buy SDK/Mobile Buy SDK/Additions/NSEntityDescription+BUYAdditions.h
@@ -44,7 +44,7 @@
 /**
  * Signifies whether this entity is private to the app (not used by the endpoint).
  */
-- (BOOL)buy_isPrivate;
+@property (nonatomic, readonly, getter=buy_isPrivate) BOOL private;
 
 /**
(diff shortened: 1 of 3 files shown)


bgulanowski:
This PR is not valid because many-to-many relationships cannot have ownership. If info is to be encoded, it should only be the ids of the destination objects. Possibly, we could always do this for many-to-many relationships, but it should only be necessary to do it in one direction.

PR #330, opened 2016-09-06
Make the sort ordering comment clearer

davidmuzi:
Make the comments for Default sort order clearer

@bgulanowski


bgulanowski:
👍

PR #331, opened 2016-09-07
Encoding many-to-many relationships in JSON

bgulanowski:
Originally, we designed the JSON encoding to ignore many-to-many relationships, because we did not have a way to specify ownership in such a relationship. Many-to-many relationships involve sharing of objects in both directions.

Any object on either side of a many-to-many may be related to multiple others. They are effectively collaborators, not parent/child. They can exist independently.

As a result, encoding a Product graph (including variants etc.) would not include the option values on the variants.

However, for the purpose of JSON encoding, some duplication may be tolerable. (Of course, if there is such duplication, then it would up to the decoder to detect it.)

One approach to avoid duplication would be to only encode the identifier. Unfortunately, in the particular case of encoding `OptionValue` (contained in `ProductVariant.options`), there _is_ no `identifier`, or any single property that can act like one. (The primary key, as such, would require two fields: `optionId` and `value`. Which is too complex a special case to address for this model.)

We never send shop catalog data across the wire back to Shopify. The only use of encoding catalog objects as JSON is for local caching. So we don't have to worry about creating duplicates in Shopify. But any clients that use this will have to be careful.

This could have had awkward consequences for the `Collection` <-> `Product` relationship, but this relationship is not described in any content returned from the listings API (which is used by the SDK to retrieve collections and products). Forming this relationship is the responsibility of client objects, determined by specifying the collection ID when requesting products. In any case, both properties have had a `JSONPropertyKey` added to their `userInfo` in the data model, so they will only encode the IDs, and then only if the relationship actually has content. 

This PR includes updates to the tests to verify this new behaviour.

@davidmuzi @gabrieloc

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYCoreDataModelAdditionsTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYCoreDataModelAdditionsTests.m
@@ -235,9 +235,24 @@ static NSString * const RootEntity = @"Root";
 
 - (void)testManyToManyRelationship
 {
-	Researcher *researcher = [self.modelManager buy_objectWithEntityName:ResearchEntity JSONDictionary:nil];
+	NSDictionary *birdJSON = @{
+							   @"id" : @1,
+							   @"colour" : @"red"
+							   };
+	NSDictionary *researcherJSON = @{
+									 @"name" : @"Bob",
+									 @"birds" : @[ birdJSON ]
+									 };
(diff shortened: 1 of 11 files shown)


bgulanowski:
Fixes #326

bgulanowski:
Closing this because time will not allow us to pursue this feature.

For future consideration, using `NSCoding` is probably a better solution for simple file caching. We could leverage the data model to avoid having to hand write `-initWithCoder:` and `-decodeWithCoder:`. Even simpler to cache JSON responses directly using NSCache or similar.

PR #332, opened 2016-09-08
Remove unused fields from Customer

### LINKED ISSUE #308: Update Customer model fields
The Customer model (and `BUYCustomer` class) include fields that are not provided by the underlying API, because they are meant to be merchant-facing.

These fields are writeable by clients:
- email: String
- acceptsMarketing: BOOL
- firstName: String
- lastName: String
- password: String
- addresses: Ordered Set

These additional fields are read-only:
- createdAt
- identifier
- defaultAddress
- lastOrderId
- lastOrderName
- multipassIdentifier
- ordersCount
- state
- taxExempt
- totalSpent
- updatedAt
- verifiedEmail

`password` is effectively write-only.

The following fields will be removed:
- note
- tags

These fields have never been provided by the API, so there is no consequence for removing them.

@davidmuzi @krisorr

bgulanowski:
This removes the `tags` and `note` fields from the `Customer` entity, which have never been available from the endpoint, and were added in error.

This does not address the apparent writability of other fields that are in fact read-only. Marking properties are read-only can be accomplished via the mogenerator template by including the key `mogenerator.readonly` in the property's `userInfo` in the managed object model definition file.

Fixes #308 

@davidmuzi


bgulanowski:
May want to cherry pick into `release/2.0.1` after this merges.

bgulanowski:
This PR is still waiting for approval.

bgulanowski:
Ping @gabrieloc

krisorr:
@bgulanowski - this is a breaking change, ie. removing functions that may have been referenced, even if they don't do anything.  Maybe it would be best to put this in 2.1.0 instead of a patch release?

bgulanowski:
Those fields were never populated and don't work. We decided it was not technically breaking.

bgulanowski:
In any case, this PR targets develop. Not serious if we don't cherry pick into 2.0.1.

krisorr:
@bgulanowski - right, my comment was only in the context of this being considered for the patch release, not merging to develop.

davidmuzi:
Technically, if its a breaking change it would merit a major version bump, 3.0.  And any commits pointing to develop will eventually make its way into a release.

Can we mark these properties as deprecated instead?

bgulanowski:
OK, sure, I'll do it next week.

PR #338, opened 2016-09-08
Bugfix/314 test errors

### LINKED ISSUE #314: sporadic crashes in unit tests
We are seeing frequent occurrences of crashes in the unit tests. Usually it's in `-[BUYRequestOperationTests testCancellationDuringExecution]`. The crash is caused by a illegal memory access exception. It happens in the testing framework.

The problem can occur locally, but is much less frequent. It could be a reference to a released object, but the crash has not been reproduced with zombies enabled, locally.

Suggest focussing test with zombies enabled on BuildKite.

@gabrieloc @davidmuzi @dbart01

### LINKED ISSUE #336: Switching test modes locally is arduous
The `test_shop_data.json` file needs to be manually edited. At best, we can stash the correct values. But they should be in version control.

Update the file with the correct values, so we don't need to submit them as a launch argument in CI. Then use a simple boolean launch argument to conditionally read those values.

bgulanowski:
Various changes to attempt to make tests faster, easier to configure, and more reliable.

Some things this PR does:

- Private data has been moved to environment variables. These can be set in a non-shared scheme, or in the CI environment, without committing to a repo.
- Configuration file has been standardized regardless whether we are using mocked responses or live server connection.
- Separate the handling of shop storefront data from configuration data:
  - break out config settings to be handled specially (domain, app id, api key, merchant id)
  - config settings can be provided in env variables (settable in scheme in Xcode)
  - config settings can alternately be provided in a file, for use in CI using same generation trick
  - the file is named `test_shop_config.json`, and exists and should be (and stay) empty
    - in any case, it is no longer tracked in version control
  - (the inability to pass env vars to `xcodebuild` to pass on to test harness is infuriating)
- Some constants have been standardized, converted from preprocessor macros.
- Removes some empty completion blocks that weren't needed
- The email address used to test sign up failure was used to ensure we get two errors from the server.

The BuildKite configuration was changed for this. There are, temporarily (until this PR is merged or rejected) two buildkite build steps: the original, run only for master and develop, and a new one run only for this branch.

## Testing Steps
- test this locally the usual way -- should work with mock data
- test this locally using environment variables:
  - clone `Mobile Buy SDK Tests` scheme
  - rename it something like `Mobile Buy SDK Tests - production`
  - make it non-shared, so it isn't saved into git
  - enter environment variables for the Test action, as follows:

clone scheme

  - run tests with the new `production` scheme selected

Still getting a failure `-testCancellationDuringPolling`, however. 😢 

Fixes #314 
Fixes #336 


Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTestBase.m
```diff
@@ -25,63 +25,83 @@
...
+NSString * const kBUYTestGiftCardCode11 = @"gift_card_code_11";
+NSString * const kBUYTestGiftCardCode25 = @"gift_card_code_25";
+NSString * const kBUYTestGiftCardCode50 = @"gift_card_code_50";
+NSString * const kBUYTestInvalidGiftCardCode = @"invalid_gift_card_code";
+NSString * const kBUYTestExpiredGiftCardCode = @"expired_gift_card_code";
+NSString * const kBUYTestExpiredGiftCardID = @"expired_gift_card_id";
+NSString * const kBUYTestDiscountCodeValid = @"discount_valid";
+NSString * const kBUYTestDiscountCodeExpired = @"discount_expired";
+NSString * const kBUYTestProductIdsCommaSeparated = @"product_ids_comma_separated";
+NSString * const kBUYTestCollectionIds = @"collection_ids";
```
dbart01: Can we make this `static NSString * const ...` instead to match the definitions for `BUYDomainName` and others? The `extern` declarations are in `BUYTestConstants.h` but the file doesn't seem to be imported anywhere, or is it?

Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
```diff
@@ -368,7 +368,7 @@ - (void)createExpectationDelay:(NSTimeInterval)delay block:(BOOL)block
 	}];
 	
 	if (block) {
-		[self waitForExpectationsWithTimeout:delay + 0.1 handler:^(NSError *error) {}];
+		[self waitForExpectationsWithTimeout:delay + 0.1 handler:[^(NSError *error) {} copy]];
```
dbart01: I imagine the the `copy` is in hope of eliminating the Buildkite crash? I'm pretty sure this accepts `nil` for the the `handler` parameter as well.

Mobile Buy SDK/Mobile Buy SDK Tests/test_shop_data.json
```diff
@@ -1,57 +1,55 @@
...
+					"2096060931",
+					"2096063363",
+					"2096062275",
+					"2096057987"
+					],
+	"collection_ids": [
+					   "109931075",
+					   "320178950",
+					   "320179206"
+					   ],
```
dbart01: This doesn't look like it was intentionally committed.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYClientTestBase.m
bgulanowski: It's imported in a few places not affected by this PR. :)

Mobile Buy SDK/Mobile Buy SDK Tests/test_shop_data.json
dbart01: Nvm, I see the reason now.

bgulanowski: They were intended, but I suppose the problem is that they are strictly tied to the private shop properties. But that's how the tests are designed. We hide the other properties because they are private and shouldn't be in version control.

bgulanowski: Oh, OK 😄

dbart01:
Looks good 🚢

bgulanowski:
I already ditched the block copy in favour of removing those empty completion blocks, but it doesn't seem to have helped.

I still think the issue is caused by a different test than the one that crashes (since the one that crashes now is not the one that used to crash before the first attempt to fix it), but still don't know.

bgulanowski:
This has been substantially updated. Sorry for the noise, but I was testing the BuildKite script.

PR #345, opened 2016-09-17
Fix params order for BUYDataTagsListBlock

sogko:
`getProductTagsPage:completion:` implementation currently passes in the params for the completion block (`BUYDataTagsListBlock`) incorrectly.

```objc
typedef void (^BUYDataTagsListBlock)(tags, page, reachedEnd, error);
```


bgulanowski:
This is already fixed by https://github.com/Shopify/mobile-buy-sdk-ios/pull/280. We will be cherry picking it into `release/2.0.1`. We don't merge PRs into master.

Issue #297, opened 2016-08-17
"Symbol not found" when compiling with xcode 8 beta, running on iOS 9


billhaley:
I'm still investigating but thought I'd run this by you and see if it rings any bells.  Trying our app with xcode 8 beta 4, and everything runs fine on iOS 10, both simulator and phone.  Running on iOS 9 immediately crashes, however. 

Console output from a crash on simulator, edited for clarity:

dyld: Symbol not found: _kABPersonAddressCityKey
  Referenced from: /Users/.../Frameworks/Buy.framework/Buy
  Expected in: .../Frameworks/PassKit.framework/PassKit
 in /Users/.../Frameworks/Buy.framework/Buy
(lldb)

bgulanowski:
`ABPersonAddressCityKey` is a symbol in AddressBook.framework. Sounds like your project or your environment is missing a reference to that framework. Are you linking to iOS 10? We haven't really tested that yet. Could be an issue with the beta tools?

davidmuzi:
`dyld: Symbol not found` errors are related to linking, where it can't resolve a symbol.  Ive just confirmed the [Advanced Sample App](https://github.com/Shopify/mobile-buy-sdk-ios/tree/master/Mobile%20Buy%20SDK%20Sample%20Apps/Advanced%20App%20-%20ObjC) runs on iOS 9.3 when compiled with iOS 10.0 SDK on Xcode 8 beta 5.  Albeit, one change was to include the Buy.framework in the Embedded Binaries section of the target.

![mobile_buy_sdk_advanced_sample_xcodeproj](https://cloud.githubusercontent.com/assets/1058176/17898311/4c811146-6925-11e6-845b-9da4121c1f7b.png)

billhaley:
@davidmuzi Thanks for looking into this!  I spent some time with it today, and the only solution that worked was the following: built the 1.2.6 framework and linked that as you've shown above.  I also needed to include both AddressBook and PassKit frameworks as well.

This does mean that using your cocoapod is no longer an option for us, which is unfortunate.  Using the pod and linking AddressBook / PassKit gives us the same error.

davidmuzi:
It looks like this issue is not limit to Cocoapods: https://openradar.appspot.com/28064734

davidmuzi:
Looks like the issue is not limited to our SDK
https://openradar.appspot.com/28064734

[...]

PR #358, opened 2016-09-29
Fix 'Symbol not found' issue with iOS 9 and Cocoapods.

wmcbain:
* Add weak qualifier to `PassKit` framework dependency in podspec.
* Resolves [#297](https://github.com/Shopify/mobile-buy-sdk-ios/issues/297)

* Additionally added `.DS_Store` to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ xcuserdata
 .ruby-version
 *Sample*/Buy.framework
 *-INTERNAL.*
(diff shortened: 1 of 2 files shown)


bgulanowski:
Thanks, we will test this.

bgulanowski:
Verified the fix. Will move the 2.0.1 tag up to include this.

Issue #355, opened 2016-09-23
Nil Checkout During `provider:didCompleteCheckout:withStatus:` Callback


cocoahero:
It appears that during a web checkout using `BUYWebCheckoutPaymentProvider`, inside of the `provider:didCompleteCheckout:withStatus:`, the provided `BUYCheckout` is nil.

We use this provided checkout for displaying an order confirmation screen, with order number, etc.

bgulanowski:
That's probably because the checkout is set to `nil` at the end of `-didReceiveCallbackURLNotification:` (line 213 or thereabouts), whereas it should happen inside the completion block provided to `-getCompletionStatusOfCheckoutURL:completion:`, invoked within.

Try moving line 213 to after line 205. I'll add this to `release/2.0.1`. Thanks for the report.

EishanVijay:
Resolved in PR #361

PR #361, opened 2016-09-30
Move nilling out of checkout to after notification is sent

davidmuzi:
Fix: #355

- Clears out the checkout object after the delegate is called

@bgulanowski


bgulanowski:
Any chance we could write a test for this?

davidmuzi:
I tried adding a unit test for this, but it requires refactoring and adding scaffolding for it to exercise it from the unit test.  After doing that, the unit test masked over the initial issue since the mocked network responses are synchronous.

Id suggest we merge this, unless you have another approach.

bgulanowski:
👍

PR #373, opened 2016-10-22
Task/enable tests

davidmuzi:
- Re-enables some disabled tests
- Corrects order of cancel before execution test

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
@@ -268,8 +268,8 @@
 	
 	[self createExpectationDelay];
 	
-	[self.queue addOperation:operation];
 	[operation cancel];
+	[self.queue addOperation:operation];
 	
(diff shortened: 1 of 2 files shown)


Mobile Buy SDK/Mobile Buy SDK Tests/BUYRequestOperationTests.m
```diff
@@ -332,7 +332,7 @@ - (void)testCancellationDuringPolling
 	
 	[self.queue addOperation:operation];
 	
-	[self createExpectationDelay:1.0 block:YES];
+	[self createExpectationDelay:10.0 block:YES];
```
bgulanowski: This adds 10 seconds to the test duration. It's not the same as waiting. Did you happen to test with any smaller values?

davidmuzi: I know.  We could be a little more aggressive, say 3.0 seconds, and back it off if we get some failures.  The 1.0s timeout was demonstrating some failures due to a race condition.

bgulanowski:
🎩

PR #377, opened 2016-10-25
Make BUYOperation header public to support Carthage integration 

nelsonbolivar:
Changed BUYOperation header from project to public
Imported BUYOperation.h into Buy.h

**Problem Description:**

When using the SDK installed with Carthage, the method `completeCheckoutWithToken` was not available, because it is returning a BUYOperation. Changing the BUYOperation to public fix the problem and the method became available.


dbart01:
The preferred solution here would be to update `completeCheckoutWithToken` to return an `NSOperation *`. All other `BUYClient` APIs return `NSOperation *` handles. We should be consistent across the SDK.

nelsonbolivar:
Using the develop branch in Carthage solve the issue with the method `completeCheckoutWithToken `. Hopefully with the next release, that problem is going to be solve.

But I'm still getting the warning saying that BUYOperation.h is not include in Buy.

untitled

bgulanowski:
The right approach is to change the return type to `NSOperation`, not to make `BUYOperation` public. `BUYOperation` is a private internal class which we do not want to expose.

We don't want to merge this change, sorry.

PR #378, opened 2016-10-26
Add minimumPrice to BUYProduct

bgulanowski:
This adds a new local-only property to BUYProduct. This enables easier sorting of products in collections. This improves parity with the Android SDK, which already has this property.

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
@@ -35,6 +35,19 @@
 
 @synthesize stringDescription=_stringDescription;
 
+- (void)setVariants:(NSOrderedSet *)variants
+{
+	[self willChangeValueForKey:BUYProductRelationships.variants];
+	NSMutableOrderedSet *primitiveVariants = [self primitiveVariants];
+	[primitiveVariants removeAllObjects];
+	if (variants.count) {
+		[primitiveVariants unionOrderedSet:variants];
+	}
+	[super setPrimitiveVariants:variants.mutableCopy];
+	[self updateMinimumPrice];
+	[self didChangeValueForKey:BUYProductRelationships.variants];
+}
+
 - (NSDate *)createdAtDate
 {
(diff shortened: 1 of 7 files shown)


Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -35,6 +35,19 @@ @implementation BUYProduct
 
 @synthesize stringDescription=_stringDescription;
 
+- (void)setVariants:(NSOrderedSet *)variants
+{
+	[self willChangeValueForKey:BUYProductRelationships.variants];
+	NSMutableOrderedSet *primitiveVariants = [self primitiveVariants];
```
davidmuzi: crashes in Advanced Sample App here.  Unrecognized selector

davidmuzi:
Should also have a unit test

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -73,6 +86,20 @@ - (NSString *)stringDescription
...
+	NSDecimalNumber *newMinimumPrice = [self.variants valueForKeyPath:@"@min.price"];
+	if (![self.minimumPrice isEqual:newMinimumPrice]) {
+		self.minimumPrice = newMinimumPrice;
+	}
+}
+
+- (void)willSave
+{
+	[super willSave];
+	[self updateMinimumPrice];
```
dbart01: `updateMinimumPrice` is called from `setVariants:` as well. Curious to know if there's ever a case where the price won't be updated if we eliminate `willSave`.

bgulanowski:
A test is failing because this inadvertently assumes the Core Data environment, which tests do not.

Closing while I address this.

bgulanowski:
In answer to @dbart01 's question, `-setVariants:` is not called when updating the object as part of deserialization. Not sure why. May only affect objects that inherit from NSObject. Will investigate that further, since that is definitely what we want, and to avoid relying on `-willSave` and `-didSave`.

Issue #376, opened 2016-10-25
property accessor intercept is wrong for relationships


bgulanowski:
Setting a relationship using `-setPrimitiveValue:forKey:` has special considerations. It requires getting the current value and updating contents, not simply replacing. As written, it will break the object and cause saves to fail.

`-setValue:forPropertyKey:` must hand ordered and unordered to-many relationships explicitly.

PR #381, opened 2016-10-26
Add minimumPrice property to BUYProduct (take 2)

bgulanowski:
This adds a new local-only property to BUYProduct. This enables easier sorting of products in collections. This improves parity with the Android SDK, which already has this property.

It might seem more natural to perform this change in the `-setVariants:` method. However, when Core Data support is enabled, `-setVariants:` is not called when updating from JSON. In any case, we don't support modifying relationships of storefront objects. This objects are read-only.

*\* Discussion **

Strictly, `-setVariants:` is not called from `-setValuesForKeysWithDictionary:`, even though it builds on `-setValue:forKey:`, which should call the custom setter. It doesn't. It is probably because the getter is not implemented, therefor KVC is not complete for that property.

In any case overriding relationship accessors for Core Data requires some sophistication. We currently have dynamic support for doing the right thing when overriding attribute accessors, but not relationship. See https://github.com/Shopify/mobile-buy-sdk-ios/issues/376

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Models/Mobile Buy SDK.xcdatamodeld/Mobile Buy SDK.xcdatamodel/contents
+++ b/Mobile Buy SDK/Mobile Buy SDK/Models/Mobile Buy SDK.xcdatamodeld/Mobile Buy SDK.xcdatamodel/contents
@@ -707,6 +707,7 @@
                 <entry key="JSONPropertyKey" value="product_id"/>
             </userInfo>
         </attribute>
+        <attribute name="minimumCompareAtPrice" optional="YES" attributeType="Decimal" defaultValueString="0.0" syncable="YES"/>
         <attribute name="minimumPrice" optional="YES" attributeType="Decimal" defaultValueString="0.0" syncable="YES"/>
         <attribute name="productType" optional="YES" attributeType="String" syncable="YES">
             <userInfo>
@@ -1006,7 +1007,7 @@
         <element name="Option" positionX="-1588" positionY="1287" width="128" height="105"/>
         <element name="OptionValue" positionX="-1390" positionY="894" width="128" height="105"/>
         <element name="Order" positionX="-549" positionY="1647" width="128" height="180"/>
-        <element name="Product" positionX="-1390" positionY="1018" width="128" height="300"/>
+        <element name="Product" positionX="-1390" positionY="1018" width="128" height="315"/>
         <element name="ProductVariant" positionX="-1192" positionY="1063" width="128" height="240"/>
         <element name="ShippingRate" positionX="-569" positionY="1836" width="128" height="135"/>
         <element name="Shop" positionX="-1786" positionY="843" width="128" height="255"/>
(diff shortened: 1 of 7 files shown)


bgulanowski:
Oh, I still have to add a unit test…

Mobile Buy SDK/Mobile Buy SDK/Models/Mobile Buy SDK.xcdatamodeld/Mobile Buy SDK.xcdatamodel/contents
```diff
@@ -321,7 +321,7 @@
         </userInfo>
     </entity>
     <entity name="Collection" representedClassName="BUYCollection" syncable="YES">
-        <attribute name="createdAt" optional="YES" attributeType="Date" syncable="YES"/>
+        <attribute name="createdAt" optional="YES" attributeType="Date" usesScalarValueType="NO" syncable="YES"/>
```
gabrieloc: Are these all automatic changes?

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.h
```diff
@@ -52,6 +52,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (NSArray<BUYProductVariant *> *)variantsArray;
 
+/**
+ * Fetch the prices from variants and cache the lowest price on minimumPrice.
+ */
+- (void)updateMinimumPrice;
```
gabrieloc: Does this need to be exposed? It feels like it should happen lazily, when `minimumPrice` is accessed. I may not have full context though.

bgulanowski: No, that would be very inefficient and expensive. We need to set it as soon as the variants as set. Lazy calculation would force us to fetch all the variants immediately, which we don't want to do when we're displaying a list of products, for example.

Mobile Buy SDK/Mobile Buy SDK/Models/Mobile Buy SDK.xcdatamodeld/Mobile Buy SDK.xcdatamodel/contents
bgulanowski: The short answer is "No", since I had to change the entity to add the property. Some of them are automated. Which ones are you particularly concerned about? All the changes to `_BUYPRoduct.[hm]` are automated.

gabrieloc: Was wondering what all the `usesScalarValueType="NO"` was about

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.h
bgulanowski: No, it doesn't have to be public. I can make it private.

Mobile Buy SDK/Mobile Buy SDK/Models/Mobile Buy SDK.xcdatamodeld/Mobile Buy SDK.xcdatamodel/contents
gabrieloc: Automatic

PR #382, opened 2016-10-26
Update checkout after authorizing Apple Pay payment

### LINKED ISSUE #362: Similar to issue #355, checkout.order is nil
Using BUYWebCheckoutPaymentProvider, after the order has been completed, "BUYPaymentProviderDidCompleteCheckoutNotificationKey" notification is not triggered. 

In addition, after checking that checkout.status == BUYStatus.complete, checkout.order is nil.  How do I get the order from checkout?

EishanVijay:
Updates checkout after Apple Pay payment is authorized

Resolves #362 

@bgulanowski @davidmuzi

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYApplePayPaymentProvider.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYApplePayPaymentProvider.m
@@ -242,7 +242,7 @@ NSString *const BUYApplePayPaymentProviderId = @"BUYApplePayPaymentProviderId";
 			default:
 				break;
 		}
-		
+		self.checkout = self.applePayAuthorizationDelegate.checkout;
 		completion(status);
 	}];
 }
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYPaymentProviderTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYPaymentProviderTests.m
@@ -29,6 +29,8 @@
 
 #import "BUYClientTestBase.h"
 #import "BUYFakeSafariController.h"
+#import "BUYApplePayAuthorizationDelegate.h"
+#import "BUYClientTestBase.h"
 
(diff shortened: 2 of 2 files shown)


Mobile Buy SDK/Mobile Buy SDK/Payment Providers/BUYApplePayPaymentProvider.m
```diff
@@ -242,7 +242,7 @@ - (void)paymentAuthorizationViewController:(PKPaymentAuthorizationViewController
 			default:
 				break;
 		}
-		
+		self.checkout = self.applePayAuthorizationDelegate.checkout;
```
davidmuzi: This change, while it looks benign, is leaking information from the delegate that the payment provider should not know about (that the checkout has changed).  A new delegate method implemented by the payment provider indicating the checkout has been updated is a possible solution.  @bgulanowski, thoughts?

bgulanowski: Then why is the checkout a public property of the authorization delegate? According to that logic, it shouldn't even be visible. Likewise some other properties, too, I would think. I believe the design exposed it as readonly because, implicitly, it was expected to change and be updated after completion.

davidmuzi: Your right, ideally it shouldnt be public.  There is no deterministic way that the PaymentProvider would know that the checkout has changed on the Delegate.  If we embrace that the checkout property is public readonly because we know it can and will be updated, KVO may be a good solution.

bgulanowski: I think we should just assume that the checkout on the delegate is always the most recent one. Moreover, the checkout on the payment provider should probably just be a cover for that of the delegate. KVO seems unnecessary when the delegate is a private implementation of the provider, and, frankly, needs to be removed in future.

PR #383, opened 2016-10-27
Update default variant check

### LINKED ISSUE #367: Default Variant check is incorrect
- The variant title is not relevant
- There should only be one optionValue
- The optionValue.title will be "Title"
- The optionValue.value will be "Default" or "Default Title" (that is currently correct)

See more info on the Android SDK issue:

https://github.com/Shopify/mobile-buy-sdk-android/issues/310

EishanVijay:
Update the default variant check to match Android and server side

Resolves #367

@bgulanowski @davidmuzi

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK Tests/BUYProductTests.m
+++ b/Mobile Buy SDK/Mobile Buy SDK Tests/BUYProductTests.m
@@ -2,19 +2,74 @@
 //  BUYProductTests.m
 //  Mobile Buy SDK
 //
-//  Created by Brent Gulanowski on 2016-10-31.
-//  Copyright © 2016 Shopify Inc. All rights reserved.
+//  Created by Shopify.
+//  Copyright (c) 2016 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
 //
 
-#import <XCTest/XCTest.h>
-
+@import XCTest;
 @import Buy;
 
 @interface BUYProductTests : XCTestCase
-
 @end
 
 @implementation BUYProductTests
+
+# pragma mark - Default Variant Checker Tests
+
+- (void)testDefaultVariantCheckerWithCorrectOptionValues {
+	BUYModelManager *model = [BUYModelManager modelManager];
+	BUYProduct *product = [[BUYProduct alloc] init];
+	BUYProductVariant *variant = [[BUYProductVariant alloc] initWithModelManager:model
+																  JSONDictionary:@{ @"id" : @1}];
+	
+	BUYOptionValue *option = [model buy_objectWithEntityName:[BUYOptionValue entityName] JSONDictionary:@{
+																										  @"name"      : @"Title",
+																										  @"value"     : @"Default",
+																										  @"option_id" : @1,
+																										  }];
+	variant.options = [[NSSet alloc] initWithArray:@[option]];
+	product.variants = [[NSOrderedSet alloc] initWithArray:@[variant]];
+	XCTAssertTrue([product isDefaultVariant]);
+}
+
+- (void)testDefaultVariantCheckerWithIncorrectOptionValues {
+	BUYModelManager *model = [BUYModelManager modelManager];
+	BUYProduct *product = [[BUYProduct alloc] init];
+	BUYProductVariant *variant = [[BUYProductVariant alloc] initWithModelManager:model
+																  JSONDictionary:@{ @"id" : @1}];
+	
+	BUYOptionValue *option = [model buy_objectWithEntityName:[BUYOptionValue entityName] JSONDictionary:@{
+																										  @"name"      : @"name1",
+																										  @"value"     : @"value1",
+																										  @"option_id" : @1,
+																										}];
+	variant.options = [[NSSet alloc] initWithArray:@[option]];
+	product.variants = [[NSOrderedSet alloc] initWithArray:@[variant]];
+	XCTAssertFalse([product isDefaultVariant]);
+}
+
+- (void)testDefaultVariantCheckerWithEmptyProduct {
+	BUYProduct *product = [[BUYProduct alloc] init];
+	XCTAssertFalse([product isDefaultVariant]);
+}
 
 - (void)testMinimumPrices
 {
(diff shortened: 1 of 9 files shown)


davidmuzi:
Can you add a unit test to demonstrate a few examples

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -118,12 +118,15 @@ - (BOOL)isDefaultVariant
...
-		NSString *defaultString = @"Default";
-		if ([productVariant.title isEqualToString:defaultTitleString] &&
-			([optionValue.value isEqualToString:defaultTitleString] || [optionValue.value isEqualToString:defaultString])) {
-			return YES;
+		if ([productVariant.options count] == 1) {
+			BUYOptionValue *optionValue = [productVariant.options anyObject];
+			NSString *defaultTitleString = @"Default Title";
+			NSString *defaultString = @"Default";
+			NSString *titleString = @"Title";
+			if ([optionValue.name isEqualToString:titleString] &&
```
bgulanowski: There's a flaw in this logic. The `optionValue.name` cannot be simultaneously equal to two different strings.

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -118,12 +118,15 @@ - (BOOL)isDefaultVariant
...
 		BUYProductVariant *productVariant = [self.variants firstObject];
-		BUYOptionValue *optionValue = [productVariant.options anyObject];
-		NSString *defaultTitleString = @"Default Title";
-		NSString *defaultString = @"Default";
-		if ([productVariant.title isEqualToString:defaultTitleString] &&
-			([optionValue.value isEqualToString:defaultTitleString] || [optionValue.value isEqualToString:defaultString])) {
-			return YES;
+		if ([productVariant.options count] == 1) {
+			BUYOptionValue *optionValue = [productVariant.options anyObject];
+			NSString *defaultTitleString = @"Default Title";
```
bgulanowski: Let's make these proper constants. Define them outside of the method scope. If they are private, make sure to declare them static:

`static NSString * const <name> = <value>;`

bgulanowski:
Is there some reason that tests aren't running on this branch?

Mobile Buy SDK/Mobile Buy SDK Tests/BUYProductTests.m
```diff
@@ -0,0 +1,74 @@
...
+@import Buy;
+
+@interface BUYProductTests : XCTestCase
+@end
+
+@implementation BUYProductTests
+
+# pragma mark - Default Variant Checker Tests
+
+- (void)testDefaultVariantCheckerWithCorrectOptionValues {
```
bgulanowski: There are two different cases, `Default` and `Default Title`. We should check for both.

You should be able to write a second test with very little effort. Factor out the common work into utility method that accepts a title value.

bgulanowski:
Please resolve conflicts.

Mobile Buy SDK/Mobile Buy SDK Tests/BUYProductTests.m
```diff
@@ -0,0 +1,72 @@
...
+	
+	BUYOptionValue *option = [model buy_objectWithEntityName:[BUYOptionValue entityName] JSONDictionary:@{
+																										  @"name"      : name,
+																										  @"value"     : value,
+																										  @"option_id" : option_id,
+																										  }];
+	variant.options = [[NSSet alloc] initWithArray:@[option]];
+	product.variants = [[NSOrderedSet alloc] initWithArray:@[variant]];
+	
+	if (valid) {
```
bgulanowski: This is good, but I suggest that the actual assertion be performed in the calling method. Passing `valid` didn't actually reduce code duplication. Instead it added complexity in the form of conditional logic.

EishanVijay:
@davidmuzi @bgulanowski Another look?

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -148,12 +152,12 @@ - (BOOL)isDefaultVariant
...
 	if ([self.variants count] == 1) {
 		BUYProductVariant *productVariant = [self.variants firstObject];
-		BUYOptionValue *optionValue = [productVariant.options anyObject];
-		NSString *defaultTitleString = @"Default Title";
-		NSString *defaultString = @"Default";
-		if ([productVariant.title isEqualToString:defaultTitleString] &&
-			([optionValue.value isEqualToString:defaultTitleString] || [optionValue.value isEqualToString:defaultString])) {
-			return YES;
+		if ([productVariant.options count] == 1) {
+			BUYOptionValue *optionValue = [productVariant.options anyObject];
```
bgulanowski: How would feel about pushing this functionality down into the child objects? It might mean we need another test or two, but I'm thinking: `-[BUYOptionValue isDefault]` and `-[BUYProductVariant isDefault]`. The former  checks the name/value of its properties. The latter checks the number of options and ensures the one is default. Then this method would check the number of variants and make sure *it* is default.

It's really not the responsibility of the product to know all this stuff about the variant and option internals.

EishanVijay: Sounds good. Would make this method a lot cleaner

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -148,12 +148,9 @@ - (BOOL)isDefaultVariant
...
 	if ([self.variants count] == 1) {
 		BUYProductVariant *productVariant = [self.variants firstObject];
-		BUYOptionValue *optionValue = [productVariant.options anyObject];
-		NSString *defaultTitleString = @"Default Title";
-		NSString *defaultString = @"Default";
-		if ([productVariant.title isEqualToString:defaultTitleString] &&
-			([optionValue.value isEqualToString:defaultTitleString] || [optionValue.value isEqualToString:defaultString])) {
-			return YES;
+		if ([productVariant isDefault]) {
+			BUYOptionValue *optionValue = [productVariant.options anyObject];
```
bgulanowski: These two lines should actually be inside `-[BUYProductVariant isDefault]`.

bgulanowski: So this method would be reducible to `return (self.variants.count == 1 && self.variants.firstObject.isDefault)`.

We should get used to using property notation.

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYOptionValue.h
```diff
@@ -31,6 +31,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (BOOL)isEqualToOptionValue:(nullable BUYOptionValue *)other;
 
+/**
+ *	Checks the name and value to determine if it's a default BUYOptionValue.
+ *
+ *	@return YES if name and value match defaults, else NO
+ */
+- (BOOL)isDefault;
```
bgulanowski: This should probably be a read-only derived property. Likewise the other properties.

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProductVariant.m
```diff
@@ -53,4 +54,8 @@ + (NSArray *)filterProductVariants:(NSArray *)productVariants forOptionValue:(BU
 	return [filteredArray copy];
 }
 
+- (BOOL)isDefault {
+	return ([self.options count] == 1 && [(BUYOptionValue *)[self.options anyObject] isDefault]);
```
bgulanowski: That cast is only necessary if you want to use the property accessor syntax.

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProduct.m
```diff
@@ -148,13 +148,7 @@ - (BOOL)isDefaultVariant
...
 	if ([self.variants count] == 1) {
 		BUYProductVariant *productVariant = [self.variants firstObject];
-		BUYOptionValue *optionValue = [productVariant.options anyObject];
-		NSString *defaultTitleString = @"Default Title";
-		NSString *defaultString = @"Default";
-		if ([productVariant.title isEqualToString:defaultTitleString] &&
-			([optionValue.value isEqualToString:defaultTitleString] || [optionValue.value isEqualToString:defaultString])) {
-			return YES;
-		}
+		return productVariant.isDefault;
```
bgulanowski: You wrote two different methods with same logic in different styles. Probably fine either way, but curious if you think one style is better than the other.

Mobile Buy SDK/Mobile Buy SDK/Models/Persistent/BUYProductVariant.m
```diff
@@ -28,6 +28,7 @@
 #import "BUYOptionValue.h"
 
 @implementation BUYProductVariant
+@synthesize isDefault;
```
EishanVijay: Project doesn't use`@derived` so I wanted to stay consistent

PR #391, opened 2016-11-01
Task/readme tests

davidmuzi:
#What
This updates the readme with correct info on running the unit tests, as well as provides an example of the config file

@bgulanowski @EishanVijay


bgulanowski:
👍

PR #396, opened 2016-11-03
[#213] Add helper method to remove discount from checkout

### LINKED ISSUE #213: Cannot remove discount after apply
We want add remove discount/promo function in our app. But we didn't find a way to do that.
I tried set `BUYDiscount` to nil and use the `updateCheckout` function in SDK, but it doesn't work. The discount is still there.

Let me know! =)

Thank you so much!

EishanVijay:
Adds a helper method to the BUYClient category to simplify the removal of a discount from the checkout.

Resolves #213 

@bgulanowski @davidmuzi @krisorr


Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.h
```diff
@@ -220,6 +220,13 @@ typedef void (^BUYDataGiftCardBlock)(BUYGiftCard * _Nullable giftCard, NSError *
  */
 - (NSOperation *)removeGiftCard:(BUYGiftCard *)giftCard fromCheckout:(BUYCheckout *)checkout completion:(BUYDataCheckoutBlock)block;
 
+/**
+ *	Removes the discount from the checkout.
+ *
+ *  @param checkout An existing BUYCheckout on Shopify
+ */
+- (void)removeDiscountFromCheckout:(BUYCheckout *)checkout;
```
EishanVijay: Android will not be adding this

davidmuzi: Then we should not add this to iOS.  cc @krisorr

bgulanowski: If we put this on `BUYCheckout` (as simply `-removeDiscount` instead of on `BUYClient`, would that be reasonable even if not on Android? I'm pretty sure we have a lot of API differences, but I understand we want `BUYClient` to be basically identical.

I actually was not expecting this to be on `BUYClient`. It's up to the app to decide when to push updates to Shopify.

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.h
```diff
@@ -220,6 +220,13 @@ typedef void (^BUYDataGiftCardBlock)(BUYGiftCard * _Nullable giftCard, NSError *
  */
 - (NSOperation *)removeGiftCard:(BUYGiftCard *)giftCard fromCheckout:(BUYCheckout *)checkout completion:(BUYDataCheckoutBlock)block;
 
+/**
+ *	Removes the discount from the checkout.
```
krisorr: nitpick but spacing is off in this comment compared to other comments in the file

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.h
krisorr: I don't see the need for a new public function for this.  On android you can do this by calling checkout.setDiscountCode(null) and doing an update call

bgulanowski: @EishanVijay let's mimic the Android approach. It's not 100% foolproof, but it's reasonable. Especially since `code` is the only property of a discount that can be set (`amount` and `applicable` are read-only).

Mobile Buy SDK/Mobile Buy SDK/Data/BUYClient+Checkout.h
EishanVijay: I think that's just Github. Looks fine in Xcode

PR #445, opened 2016-12-01
[#416] Paged View Controller for product images

### LINKED ISSUE #416: Create Product Image View Controller
When a user clicks on a product, we want to present an image view controller that shows all the images for that specific product

- [ ] page view controller for images
- [ ] placeholder image if the product has no images

EishanVijay:
Add paged view controller to show product images when a product is selected.

Created ProductItem and ImageItem classes to abstract away model objects from view controllers.

Resolves #416

@davidmuzi @bgulanowski


Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/PageViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        else {
+            return nil
+        }
+    }
+    
+    func presentationCount(for pageViewController: UIPageViewController) -> Int {
+        return productImages.count
+    }
+    
+    func presentationIndex(for pageViewController: UIPageViewController) -> Int {
```
bgulanowski: Is this method used somewhere? Can't find any references.

Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/ImageItem.swift
```diff
@@ -0,0 +1,38 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import Buy
+
+class ImageItem: NSObject {
```
bgulanowski: This is meant as a view model, I guess? Do you think this is a useful pattern in this situation? Will this class have any new properties or methods added? It's pretty empty right now.

Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/PageViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+    private func productImageViewController(forPage index: Int) -> ProductImageViewController {
+        let image = productImages[index]
+        
+        if let cachedController = productImageViewControllerCache.object(forKey: image.identifier.stringValue as NSString) {
+            return cachedController
+        } else {
+            let storyboard: UIStoryboard = UIStoryboard(name: "Main", bundle: nil)
+            guard let controller = storyboard.instantiateViewController(withIdentifier: ProductImageViewController.identifier) as? ProductImageViewController
+                else { fatalError("Unable to instantiate a ProductImageViewController.") }
+            controller.loadView()
```
bgulanowski: `loadView()` should never be called directly. Use `view` to force the view to load.

However, why do you need the view to load?

Should loading the view be implicit in the `configure(imageItem:)` method, or the `productImageView` property getter?

Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/ProductItem.swift
```diff
@@ -0,0 +1,37 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import Buy
+
+class ProductItem: NSObject {
```
bgulanowski: The name of this class specifies `Product` (as in `BUYProduct`), but it doesn't seem to know anything about products. It's better if the name of a class does not imply a specialization that does not exist. Is there a non-obvious reason why this class is only useful for product images?

Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/ViewController.swift
```diff
@@ -46,6 +46,29 @@ class ViewController: UIViewController, DataProviderSetter {
...
     }
+    
+    func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        if collectionView != self.collectionView {
+            let cell = collectionView.cellForItem(at: indexPath) as! ProductCollectionViewCell
+            if cell.productItem.productImages.count > 0 {
+                let pageViewController = PageViewController(product: cell.productItem)
+                self.present(pageViewController, animated: true, completion: nil)
+            } else {
+                let title = "Oops!"
```
bgulanowski: Suggestion: extract the alert controller creation/prep to its own method, but keep the `present()` call here.

Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/PageViewController.swift
EishanVijay: `presentationIndex(for:)` is a `UIPageViewControllerDataSource` method which returns the index of the view controller which is then reflected onto the page indicator

Mobile Buy SDK Sample Apps/Mobile Buy SDK tvOS Sample/Mobile Buy SDK tvOS Sample/ProductItem.swift
EishanVijay: This class is also used mainly to abstract model objects from views and view controllers. The class is named `ProductItem` because the class represents a product and holds the contents needed to present a product. Currently since we only present product images to the page view controller, we only need this class to store product images but the user can easily add things like product price, title, and more to this class and then use it in the view controller

gabrieloc: Same comment here as above

PR #470, opened 2016-12-15
[#468] Product Details Variant Picker

### LINKED ISSUE #468: Configure picker to show all variants in watchOS app
Configure the picker in the product details interface controller to show all the variants

EishanVijay:
Added a `WKInterfacePicker` to select variants on the product details interface controller. Only appears if the product has one option. For products with more than one option, they'll have to use their iPhone to view and buy the product.

Resolves #468

### DIFF
--- a/Mobile Buy SDK Sample Apps/Advanced App - ObjC/Mobile Buy SDK watchOS Sample App Extension/ProductItem.swift
+++ b/Mobile Buy SDK Sample Apps/Advanced App - ObjC/Mobile Buy SDK watchOS Sample App Extension/ProductItem.swift
@@ -25,18 +25,41 @@
 //
 
 import Foundation
+import Buy
+import WatchKit
 
 class ProductItem {
     
     var index: Int = 0
+    var product: BUYProduct!
     var productsModel: ProductsModel!
+    var variants: [BUYProductVariant?] = []
     
-    init(index: Int, productsModel: ProductsModel) {
+    init(index: Int, product:BUYProduct, productsModel: ProductsModel) {
         self.index = index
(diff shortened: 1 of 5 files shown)


Mobile Buy SDK Sample Apps/Advanced App - ObjC/Mobile Buy SDK watchOS Sample App Extension/ProductItem.swift
```diff
@@ -25,18 +25,54 @@
...
     func configure(controller: ProductDetailsInterfaceController) {
         self.productsModel.configure(interfaceController: controller, index: self.index)
+
+        if self.product.optionsArray().count > 1 {
+            controller.applePayButton.setHidden(true)
+            controller.variantPicker.setHidden(true)
+            controller.topSeparator.setHidden(true)
+        } else {
+            self.variants = self.product.variantsArray()
+            let caption = String(format: "%@:", (self.product.optionsArray().first?.name)!)
```
bgulanowski: Use of `optionsArray` is not guaranteed to preserve orderings. They aren't sorted, just extracted. (I'm sure it's probably deterministic to some degree, but could change across OS versions or devices.) So I'd be careful about something like `optionsArray().first?`, which does not really mean anything. (It's equivalent to `options.any?`.)

EishanVijay: Well in this case it doesn't matter since we're only going to show variants if there's only one option. If there's more than one option, the user will have to see the product on their phone.

PR #487, opened 2017-01-23
Fix the build for non-POSIX compliant shells (ie. fish)

smweber:
Don't pull the shell from $SHELL because $SHELL could be anything, but
what is really desired is a POSIX-compliant shell (ie. /bin/sh)

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK.xcodeproj/project.pbxproj
+++ b/Mobile Buy SDK/Mobile Buy SDK.xcodeproj/project.pbxproj
@@ -1913,7 +1913,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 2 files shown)


bgulanowski:
We need to update the test script to target 10.2. I don't mind if you combine that in this PR. `Mobile By SDK/Scripts/run_tests.sh`

smweber:
Looks like @davidmuzi already fixed that on master: https://github.com/Shopify/mobile-buy-sdk-ios/commit/e3cd35103cec88e96fb4de07a9f4f86ee5719acc

Want me to cherry-pick it? Or pull in master to develop? Or just let the issue take care of itself when develop is next pulled into master (my preference ;))?

bgulanowski:
I think you mean when master is pulled into develop.

You can wait if you want but we have a rule against merging PRs when tests are broken. This is a 1-character change. Whatever you think is easiest.

PR #492, opened 2017-02-09
ApplePay: Handle incomplete shipping address

glyuck:
If the user enters incomplete shipping address (i.e. without city/zip/province/country), apple pay just hangs on "Processing..." state infinitely, because `completion` callback is never called.


Mobile Buy SDK/Mobile Buy SDK/Utils/BUYApplePayAuthorizationDelegate.m
```diff
@@ -123,6 +123,9 @@ -(void)paymentAuthorizationViewController:(PKPaymentAuthorizationViewController
 	self.checkout.shippingAddress = [self buyAddressWithABRecord:address];
 	if ([self.checkout.shippingAddress isValidAddressForShippingRates]) {
 		[self updateCheckoutWithAddressCompletion:completion];
+	} else {
```
davidmuzi: This entire if statement is the same as in lines 210-215, could be extracted into a common method

PR #508, opened 2017-02-28
Sdk 3.0 Foundation

### LINKED ISSUE #507: Package generated schema and support files

dbart01:
__TLDR;__ Ignore the `+12,862 -0` stuff

### This is the beginning

SDK 3.0 starts everything from scratch and that's what we do here. There are 3 fundamental changes here:

1. Add code generation dependency - `graphql_swift_gen`.
2. Add networking layer for performing GraphQL queries / mutations - `GraphClient`, formerly known as `GraphSession`.
3. Adjust the test script to the new target, which is setup to handle new test cases as we write them.

### FAQ

- __Q__ Why is there so many file changes?
  - __A__ Ignore everything in `/Generated`. As the name implies, these are artifacts of code generation and can be safely ignored for review purposes.
- __Q__ Why is `GraphQL.swift` include in the project? Isn't is in the submodule?
  - __A__ Yes, for simplicity, it is just included in the project. In future PRs we'll leverage Swift Package Manager to handle that dependency. It's easier for now as is.
- __Q__ Why are there no tests?
  - __A__ They will be coming in future PRs. We have to tackle this in smaller portions since initial PRs are already gigantic.

Fixes #507


g-Off:
> Why is GraphQL.swift include in the project? Isn't is in the submodule?

Is this a copy of the file in `graphql_swift_gen`? Why not just have Xcode reference the file in the submodule directly instead of keeping a copy of it?

dbart01:
It is a reference, not a copy.

davidmuzi:
There has been no announcement for iOS support for SPM

dbart01:
That's correct, but there's ways to make it work and there will inevitably be official support for it in the future. For now, we're using submodules instead.

Buy/Global.swift
```diff
@@ -0,0 +1,48 @@
...
+//  Created by Dima Bart on 2017-02-24.
+//  Copyright © 2017 Shopify Inc. All rights reserved.
+//
+
+import Foundation
+
+private final class BundleCoordinator: NSObject {
+    
+    static func frameworkBundle() -> Bundle {
+        return Bundle(for: self.classForCoder())
```
bgulanowski: This is not really an appropriate use for `classForCoder()`. That is used for `NSCoding`, i.e. for encoding _instances_, not finding where a class was implemented.

Suggest you use `self.Type`.

What about making this a protocol extension on `NSObject`?

Buy/Global.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+struct Global {
```
bgulanowski: This seems like a catch-all. Why is it "global"?

Buy/GraphClient.swift
```diff
@@ -0,0 +1,147 @@
...
+        
+        precondition(!apiKey.isEmpty, "API Key is required to the Buy SDK. You can obtain one by adding a Mobile App channel here: \(shopURL.appendingPathComponent("admin"))")
+    }
+    
+    static func tokenFor(_ apiKey: String) -> String {
+        return apiKey.data(using: .utf8)!.base64EncodedString()
+    }
+    
+    static func urlFor(_ shopDomain: String) -> URL {
+        return URL(string: "https://\(shopDomain)")!
```
bgulanowski: I think there is better API for this, to prevent needing to concatenate explicitly.

Buy/Global.swift
dbart01: Honestly, it's a terrible name and I'd love a better suggestion. The intention was to provide access globally available information like bundle identifiers.

Buy/Global.swift
dbart01: You're right. Oddly enough, this is actually left over from previous experiments by accident (when it was previously an instance method). In this case, we can just use `self` instead. To take it further, we can just use `self` of `Global` and eliminate this `NSObject` subclass altogether.

g-Off:
> It is a reference, not a copy. But, ideally it shouldn't even be a reference but rather a dependency on the support Package.

Double checked, there is two copies
`mobile-buy-sdk-ios/Buy/Support/GraphQL.swift`
`mobile-buy-sdk-ios/Dependencies/Swift Gen/support/Sources/GraphQL.swift` <- should be the only copy

dbart01:
You're right, nice catch. Fixed.

PR #521, opened 2017-03-09
Graph client

dbart01:
### What this does
Remove nullability for `URLSession`. We always have one, there's no point having it declared `Optional`. Taking advantage of non-nil `URLSession`, we can also return non-nil `URLSessionDataTask`s.

### DIFF
--- a/Buy/GraphClient.swift
+++ b/Buy/GraphClient.swift
@@ -37,7 +37,7 @@ private struct Header {
 
 public class GraphClient {
     
-    public let session:  URLSession?
+    public let session:  URLSession
     
     private let apiURL:  URL
     private let headers: [String : String]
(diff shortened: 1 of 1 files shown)


g-Off:
Do you really want to start deviating like this from what we have in mobile-shopify? The reason it's an optional in mobile-shpify is because upon logout we no longer want to create tasks/have an active URLSession. Will customer accounts ever make it into this? If so then I would imagine a similar thing would be beneficial.

dbart01:
@g-Off  I'm not sure I understand how customer login state affects the underlying `GraphSession`? Customer login state is represented by a token that received / stored / deleted outside the scope of the SDK. 

Also, customer token will soon be passed as arguments into queries instead of via headers.

g-Off:
@dbart01 I had assumed it would send token via headers, but if it's going to be a query argument then that invalidates my thinking there. I think `GraphSession` will, in the near future, move to being part of the `graphql_swift_gen` submodule. We can revisit any deviations between the implementations when that happens

PR #524, opened 2017-03-09
Collections hybrid view

dbart01:
__NOTE:__ Don't get overwhelmed by the line change count. Nearly 500 lines are header comments. There's also a bunch of storyboard / nib files that you're better off looking at in Xcode.

### What this does
Adds a hybrid collection view to display collections (vertically) and products (horizontally). 

#### View models
Here, we also set the groundwork for `ViewModel` support with the 3 primary classes:
- `ViewModel.swift` - view models will implement this protocol
- `ViewModeling.swift` - objects that can be represented by a view model will implement this protocol
- `ViewModelConfigurable.swift`- objects that can be configured using a view model will implement this protocol

#### Fragments
To alleviate the pain of digging through connections & edges and to simplify query building, extensions on `Storefront` objects were introduced that provide these conveniences. These are located under the `Fragments` group in `Model` and `Query` groups respectively. They are named using the following format: `Fragment+Object.swift`.

### `StorefrontTableView` & `StorefrontCollectionView`
These subclasses offer a `cellNibName` property to Interface Builder so that a cell can be registered within Interface Builder itself.

### `UIImageView+Remote`
An extension on `UIImageView` provides asynchronous, cancellable image request functionality without relying on 3rd party dependencies like `SDWebImage` or `Alamofire`.

### DIFF
--- a/Sample Apps/Storefront/Storefront/CollectionViewModel.swift
+++ b/Sample Apps/Storefront/Storefront/CollectionViewModel.swift
@@ -13,8 +13,10 @@ struct CollectionViewModel: ViewModel {
     
     typealias ModelType = ApiSchema.Collection
     
-    let imageURL: URL?
-    let products: [ProductViewModel]
+    let title:       String
+    let description: String
+    let imageURL:    URL?
+    let products:    [ProductViewModel]
     
     // ----------------------------------
     //  MARK: - Init -
@@ -26,7 +28,9 @@ struct CollectionViewModel: ViewModel {
             self.imageURL = nil
         }
         
-        self.products = model.products().viewModels
+        self.products    = model.products().viewModels
+        self.title       = model.title
+        self.description = model.descriptionPlainSummary
     }
 }
 
--- a/Sample Apps/Storefront/Storefront/CollectionViewModel.swift
+++ b/Sample Apps/Storefront/Storefront/CollectionViewModel.swift
@@ -1,0 +1,36 @@
+//
+//  CollectionViewModel.swift
+//  Storefront
+//
+//  Created by Dima Bart on 2017-02-24.
+//  Copyright © 2017 Shopify Inc. All rights reserved.
+//
+
+import Foundation
+import Buy
+
+struct CollectionViewModel: ViewModel {
+    
+    typealias ModelType = ApiSchema.Collection
+    
+    let imageURL: URL?
+    let products: [ProductViewModel]
+    
+    // ----------------------------------
+    //  MARK: - Init -
+    //
+    init(from model: ApiSchema.Collection) {
+        if let source = model.image?.src {
+            self.imageURL = URL(string: source)!
+        } else {
+            self.imageURL = nil
+        }
+        
+        self.products = model.products().viewModels
+    }
+}
+
+extension ApiSchema.Collection: ViewModeling {
+    typealias ViewModelType = CollectionViewModel
+}
+
--- a/Sample Apps/Storefront/Storefront/CollectionViewModel.swift
+++ b/Sample Apps/Storefront/Storefront/CollectionViewModel.swift
@@ -2,8 +2,26 @@
 //  CollectionViewModel.swift
 //  Storefront
 //
-//  Created by Dima Bart on 2017-02-24.
-//  Copyright © 2017 Shopify Inc. All rights reserved.
+//  Created by Shopify.
+//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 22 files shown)


Sample Apps/Storefront/Storefront/CollectionsViewController.swift
```diff
@@ -0,0 +1,130 @@
...
+                            .products(first: 25) { $0
+                                .fragmentForStandardProduct()
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        let client = GraphClient(shopDomain: "greats.myshopify.com", apiKey: "a27a24c5570018a6908b75a3891afc5d")
```
davidmuzi: 🙅‍♂️

Sample Apps/Storefront/Storefront/UIImageView+Remote.swift
```diff
@@ -0,0 +1,81 @@
...
+
+extension UIImageView {
+    
+    typealias Completion = () -> Void
+    
+    private struct Key {
+        static var image = "com.storefront.image.key"
+    }
+    
+    private var currentTask: URLSessionDataTask? {
```
davidmuzi: Why not subclass UIImageView?

Sample Apps/Storefront/Storefront/CollectionsViewController.swift
dbart01: Not sure how this made it in. I specifically removed it. Twice.

Sample Apps/Storefront/Storefront/NSObject+Name.swift
```diff
@@ -0,0 +1,34 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+extension NSObject {
+    
+    static var className: String {
+        return NSStringFromClass(self).components(separatedBy: ".").last!
```
g-Off: `String(describing: self)`

Sample Apps/Storefront/Storefront/UIImageView+Remote.swift
dbart01: An extension on `UIImageView` allows for an easier workflow for the end user, without requiring the use of that custom subclass everywhere. This is the standard practice in libraries like `SDWebImage` and `AFNetworking`.

dbart01:
Ready for another look

Sample Apps/Storefront/Storefront/CollectionViewModel.swift
```diff
@@ -0,0 +1,58 @@
...
+    let title:       String
+    let description: String
+    let imageURL:    URL?
+    let products:    [ProductViewModel]
+    
+    // ----------------------------------
+    //  MARK: - Init -
+    //
+    init(from model: Storefront.Collection) {
+        if let source = model.image?.src {
```
g-Off: this can be cleaned up (in a subsequent PR probably) by adding
```
GraphQLSwiftGen::Scalar.new(
        type_name: 'URL',
        swift_type: 'URL',
        deserialize_expr: ->(expr) { "URL(string: #{expr})" },
        serialize_expr: ->(expr) { "#{expr}.absoluteString" },
    )
```
to the schema generation. The scalar type of URL exists in the storefront GraphQL schema and this will map it to `URL` making things a bit cleaner

Sample Apps/Storefront/Storefront/Fragment+Collection.swift
```diff
@@ -0,0 +1,35 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import Buy
+
+extension Storefront.Collection {
+    
+    func products() -> [Storefront.Product] {
```
g-Off: This would read cleaner to me as
```
var products: Storefront.Product {
    return self.products.edges.map { $0.node }
}
```

Sample Apps/Storefront/Storefront/Fragment+CollectionQuery.swift
```diff
@@ -0,0 +1,39 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
```
g-Off: Nitpick: `import Foundation` is probably not necessary, the `import Buy` is probably enough to cover everything

Sample Apps/Storefront/Storefront/CollectionViewModel.swift
dbart01: Very good point. Will update in separate PR.

Sample Apps/Storefront/Storefront/Fragment+Collection.swift
dbart01: Agreed. But, `.products` is already a defined property of `Storefront.Collection`. That's the only reason this is a method rather than a computed property. Otherwise, it'd have to be suffixed with something else.

g-Off: Ah, fair enough. I would recommend a different name then, having a function and var named the same is a tad confusing. Since this will only return the first page of products `firstProducts` or `initialProducts` might be better options (or something along those lines)

dbart01: Well not exactly. It isn't necessarily the first page. The mapping is simply from the contained edges, which can represent any page. If we say something like `after: "jdg61dn8f7"` where that value is a cursor of item on page 5, `products()` will be the 6th page. I'll rename to something more generic, though.

Sample Apps/Storefront/Storefront/CollectionViewModel.swift
dbart01: Addressed in: https://github.com/Shopify/mobile-buy-sdk-ios/pull/525

PR #534, opened 2017-03-20
Pagination

dbart01:
### What this does

Many resources in the Shopify schema are pageable but we had no support (abstractions) for easily handling cursors and page meta data in view model counterparts. We fix that here.

#### Notable changes
- introduce a `PageableArray` type that handles `pageInfo` state for any pageable collection (such as `Product`, `Image`, etc)
- use the `PageableArray` to reference any pageable collection of resources in view models. For example, `ProductViewModel`'s connection to variants is now:
```swift
let variants: PageableArray<VariantViewModel>
```
- of course, to begin paging, we'll need to know when a page ends. For this, we introduce the `Paginating` protocol that abstracts the logic of scroll view paging. Other subclasses like our custom collection and table view subclasses now implement it.
- convert view models to reference types because we need them to have an identity. Subsequent page updates rely on updating a previous page. If the view models are value types, the changes won't propagate to all references of the view model.
- remove some quirky extensions on GraphQL models that mapped `edge` `node`s since they are no longer required in creating `PageableArray` collections.


Sample Apps/Storefront/Storefront/Fragmen+VariantConnectionQuery.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  Fragmen+VariantConnectionQuery.swift
```
ishanthukral: Minor typo :)

Sample Apps/Storefront/Storefront/PaginatedArray.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  Pageable.swift
+//  Storefront
+//
+//  Created by Dima Bart on 2017-03-15.
```
davidmuzi: Use Shopify header

PR #538, opened 2017-03-21
The ParallaxViewController

dbart01:
### What this does
Adds a `ParallaxViewController`, which is a UI component that automatically manages touch events via a proxy for two overlapping views: a `UIScrollView` and an arbitrary `UIView` header view. The parallax controller also manages the layout of the header view to create a parallax effect when moving the scroll view.

### DIFF
--- a/Sample Apps/Storefront/Storefront/ParallaxViewController.swift
+++ b/Sample Apps/Storefront/Storefront/ParallaxViewController.swift
@@ -1,0 +1,135 @@
+//
+//  ParallaxViewController.swift
+//  Storefront
+//
+//  Created by Shopify.
+//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import UIKit
+
+class ParallaxViewController: UIViewController {
+    
+    @IBInspectable var insets:       UIEdgeInsets = .zero
+    @IBInspectable var headerHeight: CGFloat      = 200.0
+
+    @IBOutlet private weak var headerView: UIView!
+    @IBOutlet private weak var scrollView: UIScrollView!
+    
+    lazy private var proxyView: ProxyView = {
+        let view = ProxyView(frame: self.view.bounds)
+        self.view.addSubview(view)
+        return view
+    }()
+    
+    // ----------------------------------
+    //  MARK: - View Loading -
+    //
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        self.configureProxyView()
+        self.configureScrollView()
+    }
+    
+    private func configureProxyView() {
+        self.proxyView.headerView = headerView
+        self.proxyView.scrollView = scrollView
+    }
+    
+    private func configureScrollView() {
+        self.scrollView.backgroundColor = .clear
+    }
+    
+    // ----------------------------------
+    //  MARK: - Layout Subviews -
+    //
+    override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+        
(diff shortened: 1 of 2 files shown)


Sample Apps/Storefront/Storefront/ParallaxViewController.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+    
+    // ----------------------------------
+    //  MARK: - Updates -
+    //
+    func updateParallax() {
+        
+        let headerOffset = self.scrollView.contentOffset.y + self.midY
+        let headerY      = min(self.topY - (headerOffset * self.multiplier), self.topY)
+        let headerHeight = max(self.headerHeight, self.headerHeight + (headerOffset * -1.0))
```
davidmuzi: self.headerHeight + (headerOffset * -1.0) -> self.headerHeight - headerOffset ?

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
```diff
@@ -0,0 +1,165 @@
...
+        self.view.insertSubview(self.headerView, at: 0)
+        self.view.insertSubview(self.scrollView, at: 1)
+        self.view.insertSubview(self.proxyView, at: 2)
+    }
+    
+    // ----------------------------------
+    //  MARK: - Verification -
+    //
+    private func isChild(_ view: UIView) -> Bool {
+        for subview in self.view.subviews {
```
davidmuzi: Can't you check if the superView of the child is equal to self?

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
dbart01: lol yeah

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
dbart01: This method is actually no longer used, will remove.

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+// ----------------------------------
+//  MARK: - ProxyView -
+//
+private class ProxyView: UIView {
+    
+    weak var headerView: UIView!
+    weak var scrollView: UIScrollView!
```
g-Off: Using weak implicitly unwrapped variables seems dangerous

dbart01: I like to live on the edge.

dbart01: In all seriousness, though, it would be programmer error to use the `ParallaxViewController` with either view `nil` as it provides no value at that point. It will crash at initialization and indicate that the view controller was not configured properly. I don't see any value in gracefully failing here.

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
```diff
@@ -0,0 +1,153 @@
...
+class ParallaxViewController: UIViewController {
+    
+    @IBInspectable var insets:       UIEdgeInsets = .zero { didSet { self.view.setNeedsLayout() } }
+    @IBInspectable var headerHeight: CGFloat      = 200.0 { didSet { self.view.setNeedsLayout() } }
+    @IBInspectable var multiplier:   CGFloat      = 0.5   { didSet { self.view.setNeedsLayout() } }
+
+    @IBOutlet private weak var headerView: UIView!
+    @IBOutlet private weak var scrollView: UIScrollView!
+    
+    lazy private var proxyView: ProxyView = {
```
g-Off: Just as an FYI this style of lazy var initialization causes massive slowdowns in compile time. Every access of `.proxyView` results in having to check that the return type of the closure is the expected type. I've seen 2000+ ms times for things like this reduced to near 0 by changing approaches

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
g-Off: I think you want to use an `unowned` instead of `weak` then. Weak implies that at any point in time the object could become nil

dbart01: That's a good point. Will change.

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
dbart01: That's good to know 👍

Sample Apps/Storefront/Storefront/ParallaxViewController.swift
dbart01: Actually, an `unowned` property must have a value at initialization, which isn't possible nor desired in our case here. The values for header and scroll views are passed through from the view controller and can't be provided at init-time of `ProxyView`. Using `unowned` would require them to be set at init, making them unusable from a nib.

g-Off: If you move the construction of the `ProxyView` out of `loadView` and into `viewDidLoad` you can initialize with the needed views and have them as unowned

dbart01: The `ProxyViewController` makes no assumption about where the header and scroll view come from. For that reason, it can't ensure that both are available in `viewDidLoad` or anywhere else for that matter. They can be set at any time and can remain `nil` prior to being set. This leaves the possibility of them being set via `@IBOutlet`s or from code.

g-Off: I assume you mean the `ParallaxViewController` and not `ProxyViewController`.

Why can't it ensure that both are available in `viewDidLoad`? `viewDidLoad` is called after the views have been loaded, either via `loadView` or coming from a nib.

> They can be set at any time and can remain nil prior to being set

If that's the case then they should be weak optionals and not weak implicitly unwrapped because they could be set to being nil at any time.

dbart01: Yeah, sorry. That's what I meant. I've taken the `Optional` route here, I think that's more appropriate in this case. The `print` in the `else` conditional can also be be `fatalError` but in this case it will default to pretend the proxy view isn't there by forwarding touches to the `superview`.

PR #542, opened 2017-03-22
Product information and HTML parsing.

dbart01:
### What this does

Implements the rest of product details. Since product descriptions are provide in HTML, we needed to implement an HTML -> `NSAttributedString` converter. This class is called `HTMLParser`. It handles the styling and conversion of HTML into a useable attributed string that we display using a simple `UILabel`. Font names and sizes are injected into the HTML string using `<style></style>` tags as this is the only way to apply formatting to the provided HTML.

The product details are displayed using two types of cells:
- `ProductHeaderCell`
- `ProductDetailsCell`

### DIFF
--- a/Sample Apps/Storefront/Storefront/HTMLParser.swift
+++ b/Sample Apps/Storefront/Storefront/HTMLParser.swift
@@ -1,5 +1,5 @@
 //
-//  Fragment+ProductConnectionQuery.swift
+//  HTMLParser.swift
 //  Storefront
 //
 //  Created by Shopify.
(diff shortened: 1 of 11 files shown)


Sample Apps/Storefront/Storefront/HTMLParser.swift
```diff
@@ -0,0 +1,53 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import UIKit
+
+final class HTMLParser {
```
davidmuzi: what about an extension on String, which would be called from the html string?

dbart01: Thats probably a better alternative, will address in future PRs.

PR #551, opened 2017-03-27
Delete line items in Cart

dbart01:
### What this does

Adds ability to delete cart line items and introduces a `CartButton` that listens for updates to `Cart` and updates the badge count appropriately.


Sample Apps/Storefront/Storefront/CartButton.swift
```diff
@@ -26,38 +26,251 @@
...
+    // ----------------------------------
+    //  MARK: - Cart View -
+    //
+    private func createCartButtonView() -> CartButtonView {
         
-        let style           = NSMutableParagraphStyle()
-        style.alignment     = .center
-        style.lineBreakMode = .byClipping
+        let cartView   = CartButtonView(type: .system)
+        let cartImage  = #imageLiteral(resourceName: "cart")
```
davidmuzi: `#imageLiteral` TIL

dbart01: Funny enough, you never actually have to type out `#imageLiteral` to get auto-complete. Just typing the image name will achieve the same thing.

Sample Apps/Storefront/Storefront/SubtitleButton.swift
```diff
@@ -0,0 +1,65 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import UIKit
+
+class SubtitleButton: RoundedButton {
+    
+    @IBInspectable var subtitle: String = ""
```
davidmuzi: If the subtitle is set programmatically after title is set, it appears that it will not be rendered

dbart01: Yeah that's true. However, adding support for that would require more complicated logic since we'd need to parse the title to separate `title` from `subtitle` and update it again. Since it's not used, it's not worth the effort in my opinion.

davidmuzi:
could a naive approach be to call setTitle on didSet of subtitle ?

dbart01:
`setTitle()` requires a value for `title`, what would that be?

davidmuzi:
title :)

dbart01:
There is no `title` value that doesn't already contain the `subtitle` appended to it. This is why it's not trivial to update after the title is set.

PR #552, opened 2017-03-27
Cart button everywhere

dbart01:
### What this does

Adds the cart button to all relevant screens.

### DIFF
--- a/Sample Apps/Storefront/Storefront/CollectionsViewController.swift
+++ b/Sample Apps/Storefront/Storefront/CollectionsViewController.swift
@@ -63,6 +63,17 @@ class CollectionsViewController: UIViewController {
 }
 
 // ----------------------------------
+//  MARK: - Actions -
+//
+extension CollectionsViewController {
+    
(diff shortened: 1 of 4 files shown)


Sample Apps/Storefront/Storefront/CollectionsViewController.swift
```diff
@@ -63,6 +63,17 @@ class CollectionsViewController: UIViewController {
 }
 
 // ----------------------------------
+//  MARK: - Actions -
+//
+extension CollectionsViewController {
+    
+    @IBAction func cartAction(_ sender: Any) {
+        let cartController: CartNavigationController = self.storyboard!.instantiateViewController()
+        self.navigationController!.present(cartController, animated: true, completion: nil)
```
davidmuzi: would be safer to use the `showDetail` method on UIVIewController

dbart01: Yes, agreed. These are all addressed in later PRs.

PR #553, opened 2017-03-27
Refine cart button

dbart01:
### What this does

There are many edge cases and behaviours that affect how the cart button displays badge animations. It took some refinement to make it perfect in all cases.


Sample Apps/Storefront/Storefront/CartButton.swift
```diff
@@ -155,27 +160,84 @@ private class CartButtonView: UIButton {
...
             self.setNeedsLayout()
             self.layoutIfNeeded()
         }
         
+        guard self.badgeView.badge != badge else {
+            setter()
+            return
+        }
+        
         if animated {
```
davidmuzi: This method would be easier to read if it was broken into smaller methods

PR #556, opened 2017-03-27
3D touch

dbart01:
### What this does

Adds 3D touch capabilities for Peak & Pop to all relevant places in the sample application.


davidmuzi:
Is this needed?  While it is a cool feature, it may add unnecessary complexity to the sample app

dbart01:
It's all quite self-contained and I wouldn't say it adds that much complexity. It's only a few extensions and conformances. The architecture of the app remains simple and easy-to-understand.

PR #559, opened 2017-03-31
Apple Pay integration

dbart01:
**TLDR;** 👀  needed. If you're gonna review any PR from the SDK, this is the one to review. Also, please ignore the ludicrous `1080 additions`, a lot of is the project file and file header comments. This PR should be very reviewable.

### Intentions

The purpose of `PayController` and this abstraction is to provide the easiest solution for developers to integrate with  Pay. The intention is to limit *all* possible  Pay implementation to a reasonably-flexible, `PayCheckout` focused solution that doesn't require people to read the documentation for an hour to figure out how to use it. If you feel confused looking at this API, this is a red flag.

### What this does

Adds a target and framework called `Pay` that encapsulates everything developers will need to support Apple Pay. It is a separate target to maintain separation between the `Buy` SDK and the  Pay support. This way it's up to the developer to explicitly add a dependency on `Pay` and `PassKit`.

### Review Notes

The main class to review here is `PayController`. This is the meat of  Pay support. It provides an abstraction on top of `PKPaymentAuthorizationController` without exposing any dependencies on `PassKit`, while keeping the mental overhead to a minimum. It uses other supporting model classes like `PayCheckout` and `PayCurrency` to encapsulate data provided from `Storefront.Checkout` and other GraphQL models. 

Currently, the `PayController` doesn't provide any billing / shipping address upon successfully completing the payment. We'll need an object to encapsulate all the relevant state of a completed payment that can be used to update the checkout and send the payment token. This isn't done yet.

> Why the need for `PayCheckout` and other `Pay...` wrappers? Why not use GraphQL models directly?

Why? Since the GraphQL returns partial data and we can't guarantee the presence of fields required in `PayController`, we force end-users to create an intermediate representation of all models needed in  Pay. That way, if they forgot to request a field in their models, their code will crash (not ours).

> Where the tests at?

They're coming later.

### DIFF
--- a/Pay/PayAddress.swift
+++ b/Pay/PayAddress.swift
@@ -1,0 +1,131 @@
+//
+//  PayAddress.swift
+//  Pay
+//
+//  Created by Shopify.
+//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import PassKit
+
+public struct PayPostalAddress {
+    
+    public let city:         String?
+    public let country:      String?
+    public let countryCode:  String?
+    public let province:     String?
+    public let zip:          String?
+    
+    // ----------------------------------
+    //  MARK: - Init -
+    //
+    public init(city: String? = nil,
+         country:     String? = nil,
+         countryCode: String? = nil,
+         province:    String? = nil,
+         zip:         String? = nil) {
+        
+        self.city         = city
+        self.country      = country
+        self.countryCode  = countryCode
+        self.province     = province
+        self.zip          = zip
+    }
+}
+
+public struct PayAddress {
+    
+    public let addressLine1: String?
+    public let addressLine2: String?
+    public let city:         String?
+    public let country:      String?
+    public let countryCode:  String?
+    public let province:     String?
+    public let zip:          String?
+    
+    public let firstName:    String?
+    public let lastName:     String?
+    public let phone:        String?
+    
+    // ----------------------------------
+    //  MARK: - Init -
+    //
+    public init(addressLine1: String? = nil,
+        addressLine2: String? = nil,
+        city:         String? = nil,
+        country:      String? = nil,
+        countryCode:  String? = nil,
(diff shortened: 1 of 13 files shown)


Pay/PayCheckout.swift
```diff
@@ -0,0 +1,107 @@
...
+// ----------------------------------
+//  MARK: - PassKits -
+//
+internal extension PayCheckout {
+    
+    var summaryItems: [PKPaymentSummaryItem] {
+        var summaryItems: [PKPaymentSummaryItem] = []
+        
+        if self.hasLineItems /* or has discount */ {
+            summaryItems.append(self.lineItems.totalPrice.summaryItemNamed("CART TOTAL"))
```
davidmuzi: what about an enum, `SummaryType` which has cartTotal, subtotal, etc, and an associated value, and a function which can return a `PKSummaryItem`.

From the call site, `summaryItems.append(.total(lineItems.totalPrice).toSummaryItem()`

davidmuzi:
This PR should be broken up into smaller parts, with unit tests

Pay/PayCheckout.swift
```diff
@@ -0,0 +1,107 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import PassKit
+
+public struct PayCheckout {
```
davidmuzi: Is the idea that these structs will map 1:1 with graphQL types? Is there any way to avoid the duplicate model definitions?

Pay/PayController.swift
```diff
@@ -0,0 +1,187 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import PassKit
+
+public protocol PayControllerDelegate: class {
```
davidmuzi: where will this delegate be implemented?

Pay/PayCheckout.swift
dbart01: Good idea but since this is internal to `PayCheckout` (titles of summary items are irrelevant outside the scope of this method), I din't think it was worth the extra type declaration.

Pay/PayController.swift
dbart01: It's implemented by the owner of `PayController`. See here: https://github.com/Shopify/mobile-buy-sdk-ios/blob/sdk-3.0/sample-app/apple-pay/Sample%20Apps/Storefront/Storefront/CartViewController.swift#L170

Pay/PayCheckout.swift
dbart01: I touched on the reasons for this design in the PR description, but essentially it's to avoid `PayController` working with partially populated GraphQL models. If the needed field wasn't queried, the framework will crash, which makes debugging for end-users much more complicated.

Pay/PayAddress.swift
```diff
@@ -0,0 +1,131 @@
...
+}
+
+internal extension PayAddress {
+    
+    init(with contact: PKContact) {
+        let lines = contact.postalAddress!.street.components(separatedBy: .newlines)
+        self.init(
+            addressLine1: lines.count > 0 ? lines[0] : nil,
+            addressLine2: lines.count > 1 ? lines[1] : nil,
+            city:         contact.postalAddress!.city,
```
gavrix: `city`, `country` etc are all marked optional in initializer, might be safer to keeps using optional chaining here
```swift 
city:         contact.postalAddress?.city,
```

Pay/PayLineItem.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.price    = price
+        self.quantity = quantity
+    }
+}
+
+public extension Array where Element == PayLineItem {
+    
+    public var totalPrice: Decimal {
+        var accumulator: Decimal = 0.0
+        self.forEach { item in
```
gavrix: `reduce` might be simpler here

```
return self.reduce(Decimal(0)) { $0.price * Decimal($0.quantity) }
```

Pay/PayAddress.swift
dbart01: The `PayAddress` model doesn't make sense without a valid postal address. Agreed, this could be handled better but this functionality is augmented in future PRs.

PR #566, opened 2017-04-07
Improve checkout validation message

gavrix:
### What
Improves validation message for invalid variant_id.

Resolves https://github.com/Shopify/instant-app/issues/133
### How
In case when checkout refresh contains variant_id of the object that is not present in the channel, validation message (obtained through `-[BUYError quantityRemainingMessage]`) is ambiguous in the sense that it assumes there was issue with remaining quantity, but actual error message contains `invalid` code and no options that is assumed to be present.

Proposed change is to detect that case and show more meaningful error message for that context.

### Testing

In order to test this scenario one need to reference variant_id that is not present in the channel. I was testing on Instant-app generated for puravida shop. I altered code making checkout refresh call to following
```
- (NSOperation *)requestForURL:(NSURL *)url method:(NSString *)method 
object:(id <BUYSerializable>)object 
start:(BOOL)start completionHandler:(BUYClientRequestJSONCompletion)completionHandler
{
	NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:url];
	if (object) {
		NSDictionary *myJSON = @{
				@"checkout": @{
					@"line_items": @[
						@{
							@"variant_id":@"5620385029",
							@"quantity": @1
						}
					]
				}
		};
		
		request.HTTPBody = [NSJSONSerialization dataWithJSONObject:myJSON options:0 error:nil];
	}
...
```
and then see what error message shows in the UI
screen shot 2017-04-06 at 9 51 44 pm

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYError+BUYAdditions.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Utils/BUYError+BUYAdditions.m
@@ -72,15 +72,19 @@
 
 - (NSString *)quantityRemainingMessage
 {
-	NSNumber *remaining = (id)self.options[@"remaining"];
-	NSString *localizedString;
-	
-	if ([remaining isEqualToNumber:@0]) {
-		localizedString = NSLocalizedString(@"Completely sold out", @"String describing a line item with zero stock available");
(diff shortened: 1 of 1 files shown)


bgulanowski:
I've added @davidmuzi as a second reviewer in case he has any context we don't.

PR #569, opened 2017-04-10
Retry handling

dbart01:
### What this does

Renames and scopes client classes under `Graph` namespace. Apparently, `git` can't handle the rename well so the diff appears as a delete / add rather than a modification.

Aside from that, the point of interest in this PR is the `Graph.RetryHandler` that allows clients to provide a `condition` upon which the request will continue retrying until the `condition` returns `false`. The primary purpose of this functionality is to facilitate an easier adoption of polling APIs for fetching shipping rates and monitoring checkout status.

**NOTE: Tests are failing because they aren't running the latest version of Swift that allows nesting in generic types and other features that are required by this PR. Please disregard the failures**


Buy/Graph.RetryHandler.swift
```diff
@@ -0,0 +1,65 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+public extension Graph {
+    
+    public struct RetryHandler<R: GraphQL.AbstractResponse> {
```
davidmuzi: 🆒

Buy/Graph.Client.swift
```diff
@@ -0,0 +1,220 @@
...
+        
+        // ----------------------------------
+        //  MARK: - Init -
+        //
+        public init(shopDomain: String, apiKey: String, session: URLSession = URLSession(configuration: URLSessionConfiguration.default)) {
+            
+            let shopURL  = Client.urlFor(shopDomain)
+            self.apiURL  = Client.urlFor(shopDomain, path: "/api/graphql")
+            self.session = session
+            self.headers = [
```
davidmuzi: To be consistent with other Buy SDKs, should send:
`X-SDK-Version` which is just version, then `X-SDK-Variant` which is your free form `ios` or `android`

davidmuzi: @sav007 ^^

Buy/Graph.RetryHandler.swift
davidmuzi: should have a few unit tests

Buy/Graph.Client.swift
dbart01: Is that **in addition** to the `User-Agent` header?

davidmuzi: yea, leave it in since its indexed in Splunk so easy to query

Buy/Graph.RetryHandler.swift
dbart01: Comprehensive unit test **are** coming, just not in this PR.

davidmuzi: OK.  Prefer to have smaller PRs with Units tests though

Buy/Graph.Task.swift
```diff
@@ -0,0 +1,60 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+public extension Graph {
+    
+    public class Task {
```
davidmuzi: Whats the intended purpose of this class?

dbart01: We need reference-type wrappers for all `URLSessionDataTask`s thanks to the retry handler. If a request is retried, the original `URLSessionDataTask` will be invalid and the user will have no way to cancel retried requests. With `Graph.Task`, all handles are managed internally and any `resume` or `cancel` calls are forwarded to the relevant handle.

davidmuzi: 👍

Buy/Graph.Client.swift
dbart01: There's discussion surrounding whether this is necessary / details of implementation so I'll hold off including it in this PR.

PR #570, opened 2017-04-10
Pay session

dbart01:
### What this does

Renames `PayController` to `PaySession` and encapsulates a payment authorization in `PayAuthorization` to provide all necessary info to complete a checkout after successful  Pay invocation.

**NOTE: Disregarding failing tests, requires Xcode 8.3.***

### DIFF
--- a/Pay/PayAuthorization.swift
+++ b/Pay/PayAuthorization.swift
@@ -1,0 +1,46 @@
+//
+//  PayAuthorization.swift
+//  Pay
+//
+//  Created by Shopify.
+//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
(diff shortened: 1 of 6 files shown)


Pay/PayAuthorization.swift
```diff
@@ -0,0 +1,46 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+public class PayAuthorization {
```
davidmuzi: can this be a struct?

dbart01: Yeah, all other models are value-types. I think this was a typo.

Pay/PaySession.swift
```diff
@@ -27,33 +27,36 @@
...
+    public weak var delegate: PaySessionDelegate?
     
     public let merchantID: String
+    public let identifier: String
     
-    public fileprivate(set) var currency:      PayCurrency?
-    public fileprivate(set) var checkout:      PayCheckout?
-    public fileprivate(set) var shippingRates: [PayShippingRate]?
+    internal fileprivate(set) var currency:      PayCurrency?
+    internal fileprivate(set) var checkout:      PayCheckout?
```
davidmuzi: this property is alway being implicitly unwrapped, should be non-optional

dbart01: It can't be non-optional. A session only has a `PayCheckout` (and anything associated with one) only after a call to `authorizePaymentUsing(_: currency:)`. In theory, the checkout can also be `nil` after a payment has been completed and the session is ready to be discarded.

davidmuzi: should the property be simplicity unwrapped, then?

smweber: +1 `!`'s terrify me. Could use `guard` to unwrap?

dbart01: I don't think we should be afraid of force unwrapping. It's not inherintly "bad". In this case, we can be certain that any attempt to access `checkout` __after__ calling authorize payment method will always have a valid checkout. If doing so ever returns `nil` we most definitely want to crash because something has seriously gone wrong. An implicitly-unwrapped optional, as @davidmuzi suggested can be a more elegant solution.

smweber: So in this case you're totally right, if it's `nil` then the sky is falling and we should probably crash.

My (more general) concern is whoever will next touch this code (no idea who that might be 😉) may not check the variable declaration, assume that this implicitly-unwrapped variable is safe to use anywhere, try to access it before it's initialized, and crash. If `!`s are always avoided, the compiler will never let that happen.

That all having been said, I definitely don't feel super strongly about this. 👍

dbart01: Since `checkout` is only valid *after* starting the authorization process, this got me thinking, maybe we can just push the nullability up to the `PaySession` and more the the `checkout` and `currency` to the initializer. That way, a `PaySession` will always have an internal representation of `checkout`. Thoughts?

smweber: I don't think I have enough context with this code yet to have a strong opinion either way. I need to get digging into this project! But in general I'm in favour of anything that pushes `nil` up the tree and reduces the raw number of `!`s :D

PR #572, opened 2017-04-17
 Pay Checkout

dbart01:
### What this does

Adds  Pay checkout to the sample app using `PaySession`.

### Review notes

The majority of this PR contains no behaviour. For the most part it's mapping between models and view models and introducing more models. What does have some reviewable behaviour is the `CartViewController` that is actually responsible for  Pay checkout.

NOTE: `Client.swift` was renamed from `Graph.swift` but for some reason it shows up as an addition.

### DIFF
--- a/Sample Apps/Storefront/Storefront/AddressViewModel.swift
+++ b/Sample Apps/Storefront/Storefront/AddressViewModel.swift
@@ -1,0 +1,68 @@
+//
+//  AddressViewModel.swift
+//  Storefront
+//
+//  Created by Shopify.
+//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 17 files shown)


Sample Apps/Storefront/Storefront/CartViewController.swift
```diff
@@ -107,6 +107,34 @@ class CartViewController: ParallaxViewController {
     }
     
     // ----------------------------------
+    //  MARK: - Actions -
+    //
+    func authorizePaymentWith(_ checkout: CheckoutViewModel) {
+        let payCurrency = PayCurrency(currencyCode: "CAD", countryCode: "CA")
```
davidmuzi: should put a TODO and/or create an issue to fix this later

Sample Apps/Storefront/Storefront/CartViewController.swift
```diff
@@ -107,6 +107,34 @@ class CartViewController: ParallaxViewController {
...
+            shippingAddress: nil,
+            shippingRate:    nil,
+            discountAmount:  0.0,
+            subtotalPrice:   checkout.subtotalPrice,
+            needsShipping:   checkout.requiresShipping,
+            totalTax:        checkout.totalTax,
+            paymentDue:      checkout.paymentDue
+        )
+        
+        let paySession      = PaySession(checkout: payCheckout, currency: payCurrency, merchantID: "com.mechant.identifier")
```
davidmuzi: should be pulled from somewhere else

Sample Apps/Storefront/Storefront/AddressViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import Buy
+
+final class AddressViewModel: ViewModel {
```
davidmuzi: could this be a struct?

dbart01: All view models are reference types as they are, very often, passed around between objects (sometimes in containers). This is true for some but not all view models but for consistency I'd prefer to keep them all reference types.

g-Off: Why do they need to be reference types? It's a completely immutable data structure that's marked final and copy-on-write behaviour of Swift should give you all the benefits that I suspect you are after without requiring reference types

dbart01: While it's true that oftentimes a view model represents a dumb immutable container capable of nothing more than holding values but in this case, collection and product view models represent identity and not value. In the hybrid collection layout we have `CollectionViewModel` types that each have `products` represented by a `PageableArray<ProductViewModel>`. Each pageable array can then be individually paginated and represents the current state paging state (offset, limit, etc). If these view models were value types, updating the contents of each pageable array would involve setting new values on all instances of the collection view model. If, for example, there's more than one view controller presenting the same model on screen, the logic for finding and updating that collection would have to exist in each view controller separately. Having view models as reference types means that we can update products in a single place and have the changes propagate throughout. We want shared mutable state in this case.

PR #574, opened 2017-04-19
Graph task tests

dbart01:


BuyTests/Graph.Task.swift
```diff
@@ -0,0 +1,97 @@
+//
+//  Graph.Task.swift
```
bgulanowski: I don't understand this naming convention.

BuyTests/Graph.Task.swift
```diff
@@ -0,0 +1,97 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import XCTest
+@testable import Buy
+
+class Graph_Task: XCTestCase {
```
bgulanowski: What's wrong with `TaskTest`? "Graph" seems verbose. It's implicit.

BuyTests/Graph.Task.swift
dbart01: `Task` is a nested class under the `Graph` namespace. It neatly groups all classes pertaining to GraphQL under one namespace.

BuyTests/Graph.Task.swift
dbart01: This name is incorrect and is fixed in future PRs.

PR #589, opened 2017-04-24
Buildkite pipeline

dbart01:
### The problem

Previously, the SDK build process involved a single script that would "run tests". The path to this script along with any required environment variables were hard-coded in Buildkite admin making it very difficult (impossible) to have custom build environments on different branches.

Also, in SDK 3.0 we have two independent targets that needs testing. With a single script, we can't parallelize the builds.

### Solution

Migrate to using a `pipeline.yml` file that would setup the build pipeline dynamically. Doing so, allows us to setup steps and environment as needed on any specific branch, as well as, run multiple target's tests in parallel.

### Notes

Adding the `pipeline.yml` required an addition hard-coded step in Buildkite to ensure backwards compatibility with the previous method. It would upload the pipeline if one was found, or continue test execution as it always has in the past. This step looks like this:

screen shot 2017-04-24 at 9 05 48 am

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,17 +1,15 @@
+env:
+  TARGET_IMAGE: "macos-xcode83"
+  
 steps:
-
   - label: "Buy Tests"
     command: "./Scripts/test_buy"
     timeout_in_minutes: 10
     agents:
       macstadium-scheduler: true
-    env:
-      TARGET_IMAGE: "macos-xcode83"
     
   - label: "Pay Tests"
     command: "./Scripts/test_pay"
     timeout_in_minutes: 10
     agents:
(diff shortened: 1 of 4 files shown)


davidmuzi:
CircleCI should be considered, as it will give contributors outside Shopify to view the results.  @sander-m

dbart01:
We can also use TravisCI, but I've heard it's slower than CircleCI in the free, open-source tier.

sander-m:
👍

davidmuzi:
@sander-m whats the preferred CI system to use?

sander-m:
Circle or Travis for open source projects.

dbart01:
Done: https://github.com/Shopify/mobile-buy-sdk-ios/pull/590

PR #593, opened 2017-05-01
Card vaulting

dbart01:
It's like pole vaulting but with credit cards.

### What this does

Adds support for card vaulting. Introduces a few new classes to make this happen:
- The `Card.Client` is like `Graph.Client` but only for card server
- The `Card.CreditCard` model encapsulates all credit card data and provides an internal dictionary serialization implementation
- Tests with 100% coverage for all the above

### DIFF
--- a/Buy/Card.Client.swift
+++ b/Buy/Card.Client.swift
@@ -79,7 +79,8 @@ public extension Card {
         /// that can be used to complete a checkout.
         ///
         /// - parameters:
-        ///     - creditCard: The credit card to Vaulting
+        ///     - creditCard: The credit card to vault
+        ///     - url:        The `vaultUrl` obtained from a `Storefront.Checkout`
         ///     - completion: The completion handler that will be executed with the token after a card has has been vaulted.
         ///
         /// - returns:
--- a/Buy/Card.Client.swift
+++ b/Buy/Card.Client.swift
@@ -44,8 +44,6 @@ public extension Card {
         /// The `URLSession` backing all `Client` network operations. You may provide your own session when initializing a new `Client`.
         public let session: URLSession
         
-        internal let apiURL: URL = URL(string: "https://elb.deposit.shopifycs.com/sessions")!
-        
         // ----------------------------------
         //  MARK: - Init -
         //
@@ -61,8 +59,8 @@ public extension Card {
         // ----------------------------------
         //  MARK: - Request Generation -
         //
-        func requestFor(_ creditCard: CreditCard) -> URLRequest {
-            var request = URLRequest(url: self.apiURL)
+        func requestFor(_ creditCard: CreditCard, to url: URL) -> URLRequest {
+            var request = URLRequest(url: url)
             
             request.httpMethod = "POST"
             request.httpBody   = try! JSONSerialization.data(withJSONObject: creditCard.dictionary(), options: [])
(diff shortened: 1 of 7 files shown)


Buy/Card.Client.swift
```diff
@@ -0,0 +1,112 @@
...
+    /// The `Vault.Client` is a network layer designed to abstract the communication with the Shopify's
+    /// credit card server. For PCI-compliance reasons, this is a standalone service, separate from the
+    /// main `Graph.Client` service.
+    ///
+    public class Client {
+     
+        /// The `URLSession` backing all `Client` network operations. You may provide your own session when initializing a new `Client`.
+        public let session: URLSession
+        
+        internal let apiURL: URL = URL(string: "https://elb.deposit.shopifycs.com/sessions")!
```
davidmuzi: Is this always the same URL? With the old Checkout API, the vault URL would be part of the checkout object

Buy/Card.swift
```diff
@@ -0,0 +1,29 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+public struct Card {}
```
davidmuzi: why empty class definition + extension?

Buy/Card.Client.swift
dbart01: As per Shopify docs [here](https://help.shopify.com/api/reference/checkout#create_payment_session):
```
The card vault has a static url and is located at https://elb.deposit.shopifycs.com/sessions.
```

Buy/Card.swift
dbart01: It's acting as a namespace.

Buy/Card.Client.swift
davidmuzi: Thats the admin endpoint.  Are you sure its the same one for Storefront? @maximebedard

Buy/Card.swift
davidmuzi: Right - but having the implementation inside the class, in one file, accomplished the same thing

Buy/Card.Client.swift
maximebedard: It's the same endpoint (PCI compliant endpoint completely outside of shopify/shopify)

Buy/Card.swift
dbart01: Sure, but that pattern doesn't scale. We use exactly the same approach with the `Graph` namespace. The pattern is designed to explicitly declare a namespace - in a separate file. And provide extensions to that namespace for any type that should belong to it.

Buy/Card.Client.swift
dbart01: Should we query the `vaultUrl` on `Checkout` instead actually?

maximebedard: I'd say yes, but technically it's always the same 😄

Buy/Card.swift
davidmuzi: My bad, I glanced over the `CreditCard` struck _inside_ `Card` 👍

PR #595, opened 2017-05-04
Cache

dbart01:
### What this does

Adds `Graph.Cache` and `Graph.CacheItem` to facilitate caching responses to disk using the query as the key. The `Graph.Client` doesn't actually do anything with the cache yet. That functionality is coming in a later PR. For now, it's just the cache classes and unit tests.

#### How it works

A `Graph.CacheItem` represents the data of a cacheable response along with a timestamp and hash. Each `Graph.CacheItem` is stored as two separate files on disk:

- `37d3868e50b398dc12ddd14ba1cec315`
- `37d3868e50b398dc12ddd14ba1cec315.meta`

Where the plain hash file contains the response data and the `.meta` file contains a JSON of the remaining meta data, like this:
```json
{
  "hash": "37d3868e50b398dc12ddd14ba1cec315",
  "timestamp": 1493923948,
}
```

The `Graph.Cache` is then able to write and read cached data corresponding to the `Hash` passed in as the key. To speed up reads and writes and avoid hitting the disk for every cache access, `Graph.Cache` also implements an `NSCache` to store the responses in memory. `NSCache` will then automatically purge objects based on memory pressure.

#### What's up with the weird Crypto target?

Since `libCommonCrypto.dylib` is a C-based library and not a module, Swift has a really hard time figuring out what it is. Ideally, we just want to `import Crypto` and go on our merry way. To do this, we introduce a new target with a custom `.modulemap` file that simply maps the `CommonCrypto` header to a module. Having done that, we **can** `import Crypto` as any other framework 🎉

### DIFF
--- a/Buy/Digest.swift
+++ b/Buy/Digest.swift
@@ -1,0 +1,60 @@
+//
+//  Digest.swift
+//  Buy
+//
+//  Created by Shopify.
+//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import Crypto
+
+struct Digest {
+    
+    static func md5(_ data: Data) -> [UInt8] {
+        
+        var context = CC_MD5_CTX()
+        var digest  = [UInt8](repeating: 0, count: Int(CC_MD5_DIGEST_LENGTH))
+        
+        CC_MD5_Init(&context)
+        data.forEach { byte in
+            CC_MD5_Update(&context, [byte], 1)
+        }
+        CC_MD5_Final(&digest, &context)
+        
+        return digest
+    }
+}
+
+extension String {
+    
+    var md5: String {
+        let data = self.data(using: .utf8)!
+        return Digest.md5(data).hex
+    }
+}
+
+extension Array where Element == UInt8 {
+    
+    var hex: String {
+        return self.map { String(format: "%02x", $0) }.joined()
+    }
+}
--- a/Buy/Digest.swift
+++ b/Buy/Digest.swift
@@ -47,8 +47,14 @@ struct Digest {
 extension String {
     
     var md5: String {
-        let data = self.data(using: .utf8)!
-        return Digest.md5(data).hex
+        return self.data(using: .utf8)!.md5
+    }
+}
+
+extension Data {
+    
+    var md5: String {
+        return Digest.md5(self).hex
     }
 }
 
--- a/Buy/Digest.swift
+++ b/Buy/Digest.swift
@@ -29,6 +29,8 @@ import Crypto
 
 struct Digest {
     
+    private init() {}
+    
     static func md5(_ data: Data) -> [UInt8] {
         
         var context = CC_MD5_CTX()
--- a/Buy/Graph.CacheItem.swift
+++ b/Buy/Graph.CacheItem.swift
@@ -90,6 +90,9 @@ internal extension Graph {
     }
 }
 
+// ----------------------------------
+//  MARK: - Location -
+//
 extension Graph.CacheItem {
     internal struct Location {
         
@@ -111,5 +114,19 @@ extension Graph.CacheItem {
             self.dataURL = dataURL
             self.metaURL = metaURL
         }
+    }
+}
+
+// ----------------------------------
+//  MARK: - URLRequest Hash -
+//
+extension URLRequest {
+    
+    var cacheItem: Graph.CacheItem {
+        return Graph.CacheItem(hash: self.hash, data: self.httpBody ?? Data())
+    }
+    
+    var hash: Graph.Hash {
+        return (self.httpBody ?? Data()).md5
     }
 }
--- a/Buy/Graph.CacheItem.swift
+++ b/Buy/Graph.CacheItem.swift
@@ -110,7 +110,7 @@ extension Graph.CacheItem {
             )
         }
         
-        init(dataURL: URL, metaURL: URL) {
+        internal init(dataURL: URL, metaURL: URL) {
             self.dataURL = dataURL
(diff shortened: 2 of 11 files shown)


Buy/Graph.CachePolicy.swift
```diff
@@ -0,0 +1,50 @@
...
+        case cacheOnly
+        
+        /// Load from network without loading from cache
+        case networkOnly
+        
+        /// Load from cache if staleness interval is not exceeded, otherwise load from network
+        case cacheFirst(expireIn: Int)
+        
+        /// Load from network but fallback to cached data if the request fails
+        case networkFirst(expireIn: Int)
```
davidmuzi: should a cacheThenNetwork be supported?

BuyTests/Graph.CacheItemTests.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+    
+    // ----------------------------------
+    //  MARK: - Write -
+    //
+    func testWrite() {
+        let item     = self.defaultCacheItem()
+        let location = Graph.CacheItem.Location(inParent: self.testDirectory, hash: "test-item")
+        
+        XCTAssertFalse(self.fileManager.fileExists(atPath: location.dataURL.path))
```
davidmuzi: Im seeing double 👀

BuyTests/Graph.CacheItemTests.swift
```diff
@@ -0,0 +1,142 @@
...
+        XCTAssertFalse(self.fileManager.fileExists(atPath: location.dataURL.path))
+        
+        let success = item.write(to: location)
+        
+        XCTAssertTrue(success)
+        XCTAssertTrue(self.fileManager.fileExists(atPath: location.dataURL.path))
+        XCTAssertTrue(self.fileManager.fileExists(atPath: location.dataURL.path))
```
davidmuzi: here too

Buy/Graph.CacheItem.swift
```diff
@@ -0,0 +1,132 @@
...
+            self.timestamp = result.meta["timestamp"] as! Double
+        }
+        
+        // ----------------------------------
+        //  MARK: - IO -
+        //
+        private static func read(from location: Location) -> (data: Data, meta: [String : Any])? {
+            guard let data = try? Data(contentsOf: location.dataURL),
+                let meta = try? Data(contentsOf: location.metaURL),
+                let json = (try? JSONSerialization.jsonObject(with: meta, options: [])) as? [String : Any] else {
```
davidmuzi: Curious, why JSON over NSCoding?

dbart01: `NSCoding` and Swift really don't mix well. Also I didn't want `CacheItem` to inherit from `NSObject`.

BuyTests/Graph.CacheItemTests.swift
dbart01: One of them should be `metaURL`, nice catch 👍

Buy/Graph.CachePolicy.swift
dbart01: That's what `cacheFirst` is, no? Look into cache, if expired then fetch network.

Buy/Graph.CacheItem.swift
davidmuzi: neither does JSON and Swift for that matter :)

dbart01:
> Can you explain how this will be used? How/where will staleness be checked, and how will old items be deleted?

I was hoping to save that for the next PR that will actually use it. But in summary, the `Graph.Client` evaluates the staleness based on the stored timestamp. If it exceed the `expireIn` interval, the item is either purged or simply left untouched and the network is fetched instead. All cache items are stored in the `tmp` where the system will eventually clean up any left over items.

Buy/Graph.CachePolicy.swift
davidmuzi: The comment suggests Cache, otherwise Network (1 or the other)

dbart01: What's the logic in `cacheThenNetwork`?

davidmuzi: returns cache for immediate response, and then goes to the network too, to get any updated results

Buy/Graph.CacheItem.swift
dbart01: Fair enough.

Buy/Graph.CachePolicy.swift
dbart01: We originally intended that behaviour for `cacheFirst` but after some discussion we realized that multiple-callback behaviour doesn't work with synchronous client requests on Android. As a result, to stay consistent across platforms we opted for the current implementation.

cc: @sav007

Buy/Graph.CacheItem.swift
gavrix: There are good chances JSON and swift relationship will get upgrade in near future though

Buy/Graph.CacheItem.swift
```diff
@@ -0,0 +1,132 @@
...
+        //
+        init(inParent url: URL, hash: Graph.Hash) {
+            let dataURL = url.appendingPathComponent(hash)
+            self.init(
+                dataURL: dataURL,
+                metaURL: dataURL.appendingPathExtension("meta")
+            )
+        }
+        
+        init(dataURL: URL, metaURL: URL) {
```
gavrix: Looks like this might have been private, to ensure  `metaURL` = `dataURL` + `.meta`

Buy/Graph.CachePolicy.swift
gavrix: Each request should deliver single response. I'd say for multiple results it should be a combination of 2 requests — one explicitly form cache, the other — explicitly from network.

Buy/Graph.CacheItem.swift
```diff
@@ -0,0 +1,132 @@
...
+                let json = (try? JSONSerialization.jsonObject(with: meta, options: [])) as? [String : Any] else {
+                    
+                    return nil
+            }
+            
+            return (data, json)
+        }
+        
+        @discardableResult
+        func write(to location: Location) -> Bool {
```
gavrix: Curious, why not re-throwing underlying error?

Buy/Digest.swift
```diff
@@ -0,0 +1,66 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import Crypto
+
+struct Digest {
```
gavrix: It is possible to create useless instance of this struct, which doesn't seem to make sense in this design. Having `Digest` as empty `enum` (no constructor) of `final class` or `struct` with unavailable constructor might have been clearer in terms of API.

Buy/Graph.CacheItem.swift
dbart01: Mostly because the error, if occurs, really provides no value to the caller. It should almost never fail and if it does, the result of the operation is not critical. One case I can think of is if there's not enough disk space, in which case, caching is not top priority.

Buy/Graph.CacheItem.swift
dbart01: That doesn't necessarily have to be enforced since those locations can be different. The `metaURL = dataURL + .meta` condition is only enforced by the other initialer that explicitly takes a `parentDirectory` as a parameter. Having said that, it doesn't hurt to make this `private`.

Buy/Digest.swift
dbart01: Agreed. Will adjust.

Buy/Graph.CachePolicy.swift
dbart01: Correct me if I'm wrong but right now a single request delivers a single response. For `cacheFirst(expireIn:)` this is the flow:
1. check if response is in the cache
    - Yes, go to `#2`
    - No, got to `#3`
2. Check if response expiry + expireIn > now
    - Yes, got to `#4`
    - No, go to `#3`
3. Fetch network
4. Return response

The callback is only ever called once.

dbart01:
Ready for another 👀

dbart01:
Ignore failing Buildkite builds, CircleCI is the source of truth.

PR #597, opened 2017-05-12
Implement Cache in Graph.Task

dbart01:
### What this does

Since `URLSession` doesn't allow us to inject some kind interceptor for network requests in order to customize caching behaviour, we are force to wrap the `URLSession` functionality in a custom task to proxy calls to our `Cache` and `URLSession` as needed. Some notable changes include:
- convert `Graph.Task` to a protocol an reimplement the session behaviour in a backing `Graph.InternalTask` rather than the `Graph.Client`
- implement 4 caching behaviours as dictated by `Graph.CachePolicy`
- rewrite / update unit tests for `Client` and `InternalTask` accordingly


Buy/Graph.Task.swift
```diff
@@ -26,42 +26,271 @@
...
+        
+        private func resume(using cachePolicy: CachePolicy) {
+            
+            let request    = self.request
+            let completion = self.completion
+            let hash       = request.hash
+            
+            switch cachePolicy {
+            case .cacheFirst(let expireIn):
+                print("Exercising cache policy: CACHE_FIRST(\(expireIn))")
```
davidmuzi: Are these prints intended for release too?

Buy/Graph.Task.swift
```diff
@@ -26,42 +26,271 @@
...
+                self.task = self.graphTaskWith(request, retryHandler: self.retryHandler) { response, data, error in
+                    
+                    if let _ = response, let data = data {
+                        self.cache(data, for: hash)
+                        completion(response, error)
+                        
+                    } else {
+                        
+                        self.cachedModelFor(hash, expireIn: expireIn) { response in
+                            if let response = response {
```
davidmuzi: If both response and error are optional, why not just call the completion?

Buy/Graph.Client.swift
```diff
@@ -151,56 +150,41 @@ extension Graph {
...
-            request.httpMethod = "POST"
-            request.httpBody = String(describing: query).data(using: .utf8)
+            request.httpMethod              = "POST"
+            request.httpBody                = requestData
             request.httpShouldHandleCookies = false
-            request.setValue("application/json", forHTTPHeaderField: Header.accept)
+            
+            request.setValue("application/json",    forHTTPHeaderField: Header.accept)
             request.setValue("application/graphql", forHTTPHeaderField: Header.contentType)
+            request.setValue(requestData.md5,       forHTTPHeaderField: Header.queryTag)
```
davidmuzi: Is this something the server requires?  If its just for the client, why not calculate the hash from the requests payload when its needed?

Buy/Graph.Task.swift
```diff
@@ -26,42 +26,271 @@
...
-            self.task = task
+        func resume() {
+            self.isResumed = true
+            self.resume(using: self.cachePolicy)
+        }
+        
+        private func resume(using cachePolicy: CachePolicy) {
+            
+            let request    = self.request
+            let completion = self.completion
```
davidmuzi: what are these local variable declarations for?

davidmuzi: oh, recursion

Buy/Graph.Task.swift
```diff
@@ -26,42 +26,271 @@
...
+            let hash       = request.hash
+            
+            switch cachePolicy {
+            case .cacheFirst(let expireIn):
```
davidmuzi: Have you explored extracting this behaviour into the enum, or another struct to localize the implementation and make it easier to test?

Buy/Graph.Task.swift
```diff
@@ -26,42 +26,271 @@
...
+        
+        // ----------------------------------
+        //  MARK: - Session -
+        //
+        private func graphTaskWith(_ request: URLRequest, retryHandler: RetryHandler<R>? = nil, completion: @escaping (R?, Data?, QueryError?) -> Void) -> URLSessionDataTask {
+            return self.session.dataTask(with: request) { data, response, error in
+                
+                let (model, error) = self.processResponse(data, response, error)
+                
+                DispatchQueue.main.async {
```
davidmuzi: Should you rely on the callback `delegateQueue` set on the URLSession instead?

Buy/Graph.Client.swift
dbart01: It's not strictly necessary but it makes subsequent `hash` calculations much faster - `O(1)` vs `O(n)` where `n` is the size of the HTTP body in bytes. It's also very nice to have for debugging purposes both on client and server to be able to distinguish identical queries. It doesn't hurt anything to have this header in place.

Buy/Graph.Task.swift
dbart01: These are leftover from a previous implementation. They were used to avoid capturing `self` in closures below. `request` is no longer used and should be replaced with `cache` instead. The `hash` local variable is to prevent calculating the the hash more than once.

Buy/Graph.Task.swift
dbart01: I'm not sure I understand what behaviour you mean. `.cacheFirst` is an `enum`, no?

Buy/Graph.Task.swift
dbart01: This needs to be executed on the main thread, not the session callback thread. The returned `Task` is not thread-safe (nor is the `Client` for that matter). Anything we return from `Client` is assumed to be on the main thread. Any mutations to `Task`, then, should also be on the main thread.

Buy/Graph.Task.swift
dbart01: Because we have hit the cache only after a network error, which means the `error` is always not `nil` in this scope. We don't want to return the error if the cache has a response.

Buy/Graph.Task.swift
dbart01: Nope, this is fixed here #598

Buy/Graph.Task.swift
davidmuzi: right, you could have the methods implemented on the enum itself, could make testing easier

dbart01: I'm not sure I see the benefit of making the `CachePolicy` any smarter than it currently is. Managing the request flow isn't the responsibility of the `CachePolicy`. It shouldn't know anything beyond simple enum values. Otherwise it'll need to know about the `client`, `cache`, `retryHandler`, etc.

PR #602, opened 2017-05-23
Client cache policy

dbart01:
### What this does

Adds a default client-wide `CachePolicy` that will be extended to each `query` operation if no cache policy is provided. Add tests for this behaviour.

### DIFF
--- a/Buy/Graph.Client.swift
+++ b/Buy/Graph.Client.swift
@@ -59,6 +59,9 @@ extension Graph {
     ///
     public class Client {
         
+        /// Cache policy to use for all `query` operations produced by this instance of `Graph.Client` 
+        public var cachePolicy: CachePolicy = .networkOnly
+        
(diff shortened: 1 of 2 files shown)


Buy/Graph.Client.swift
```diff
@@ -59,6 +59,9 @@ extension Graph {
     ///
     public class Client {
         
+        /// Cache policy to use for all `query` operations produced by this instance of `Graph.Client` 
+        public var cachePolicy: CachePolicy = .networkOnly
```
davidmuzi: would this ever change at runtime? Why not a let and set the default value in the constructor?

dbart01: Constructors should only require parameters that are essential for initialization, which a caching policy isn't. We can't predict what the use case might be so it's left as a mutable property. Is there any particular reason why you think this shouldn't be mutable?

PR #614, opened 2017-05-31
remove variant from cart

BKubavat:
Shopify is already giving option to decrease the quantity. If there would be product quantity of three now it would be easy to delete product from cart only by one call.


BKubavat:
So where am I missing in deleting product from cart directly?

PR #637, opened 2017-06-12
Crypto

### LINKED ISSUE #635: Framework does not build on systems that don't have /Applications/Xcode.app
The `Crypto/iphone.modulemap` file points explicitly to `/Applications/Xcode.app`. It is unreasonable to require Xcode be named anything specific for the project to build. There is little benefit from talking to CommonCrypto directly from Swift when a couple of well-annotated Objective-C functions will do just fine.

asdf

dbart01:
### What this does

- removes the custom target that allowed us to point to `libCommonCrypto`
- use a c implementation of the `md5` hash instead

This paves the way for CocoaPods support.

Fixes #635


davidmuzi:
Will this address https://github.com/Shopify/mobile-buy-sdk-ios/issues/635

dbart01:
Yes it will fix #635, updated description.

calebd:
Curious, why not a C function that calls into CommonCrypto? You can still use it from a C or Objective-C function without needing a custom Swift module or needing to provide your own implementation.

dbart01:
@calebd That's a good idea. This change will be included in the next release #642. Thanks for the contribution 👍

PR #656, opened 2017-06-22
Schema update

dbart01:
### What this does

Runs the automated update routine and introduces the following changes to the generated schema:
- reduce supported query parameters for `Customer.orders`
- add `phone` parameter to `CustomerCreateInput`
- add `phone` parameter to `CustomerUpdateInput`
- add `formattedArea` to `MailingAddress`
- reduce sort keys exposed for customer orders
- add `availableForSale` instead of `available` in `ProductVariant`
- add ability to fetch a collection or product by handle

Will need similar on Android: @sav007

### DIFF
--- a/Buy/Generated/Storefront/OrderSortKeys.swift
+++ b/Buy/Generated/Storefront/OrderSortKeys.swift
@@ -29,25 +29,13 @@ import Foundation
 extension Storefront {
 	/// The set of valid sort keys for the orders query. 
 	public enum OrderSortKeys: String {
-		case createdAt = "CREATED_AT"
-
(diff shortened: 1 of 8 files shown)


Buy/Generated/Storefront/OrderSortKeys.swift
```diff
@@ -29,26 +29,14 @@ import Foundation
 extension Storefront {
 	/// The set of valid sort keys for the orders query. 
 	public enum OrderSortKeys: String {
-		case createdAt = "CREATED_AT"
```
davidmuzi: Are these not breaking changes?  Why are these not marked deprecated?

PR #677, opened 2017-07-05
Fix date initialization for 12hour time format in UTC-0 region.

### LINKED ISSUE #665: Incorrect api behavior
Swift, iOS, SDK version 2.2.0

I use property `publishedAtDate` in `BUYProduct` object. And in your code it is marked as NonNull and for some users it crashes.
```
libswiftFoundation.dylib       0x10148c3d0 static
Date._unconditionallyBridgeFromObjectiveC(NSDate?) -> Date (__hidden#3516_:261)
```

My assumption is publishedAtDate in swift code has type Date, not Date? or Date!. That's why I can't safety unwrap this property. And when SDK sends me nil code crashes.

ucerxxx:
Fixes date initialization fail for case when user locale is in UTC-0 and time format is set as non 24-hour.

Date formatter locale is set to "us_GB" instead of currentLocale if 12-hour time format is set in user settings.

Fixes #665

### DIFF
--- a/Mobile Buy SDK/Mobile Buy SDK/Additions/NSDateFormatter+BUYAdditions.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Additions/NSDateFormatter+BUYAdditions.m
@@ -38,29 +38,9 @@
 + (NSDateFormatter*)dateFormatterForPublications
 {
 	NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
-	
-	if (![self is24hTimeFormat]) {
-		[dateFormatter setLocale:[NSLocale localeWithLocaleIdentifier:@"en_GB"]];
-	}
-	
+	[dateFormatter setLocale:[NSLocale localeWithLocaleIdentifier:@"en_GB"]];
 	dateFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ssZ";
 	return dateFormatter;
-}
-
-+ (BOOL)is24hTimeFormat
-{
-	NSDateFormatter *formatter = [[NSDateFormatter alloc] init];
-	[formatter setLocale:[NSLocale currentLocale]];
-	[formatter setDateStyle:NSDateFormatterNoStyle];
-	[formatter setTimeStyle:NSDateFormatterShortStyle];
-	
-	NSString *dateString = [formatter stringFromDate:[NSDate date]];
-	
-	NSRange amRange = [dateString rangeOfString:[formatter AMSymbol]];
-	NSRange pmRange = [dateString rangeOfString:[formatter PMSymbol]];
-	BOOL is24h = (amRange.location == NSNotFound && pmRange.location == NSNotFound);
-	
-	return is24h;
 }
 
 @end
--- a/Mobile Buy SDK/Mobile Buy SDK/Additions/NSDateFormatter+BUYAdditions.m
+++ b/Mobile Buy SDK/Mobile Buy SDK/Additions/NSDateFormatter+BUYAdditions.m
@@ -38,8 +38,29 @@
 + (NSDateFormatter*)dateFormatterForPublications
 {
 	NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
+	
+	if (![self is24hTimeFormat]) {
+		[dateFormatter setLocale:[NSLocale localeWithLocaleIdentifier:@"en_GB"]];
+	}
+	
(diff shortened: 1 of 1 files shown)


Mobile Buy SDK/Mobile Buy SDK/Additions/NSDateFormatter+BUYAdditions.m
```diff
@@ -38,8 +38,29 @@ + (NSDateFormatter*)dateFormatterForShippingRates
...
+	[formatter setDateStyle:NSDateFormatterNoStyle];
+	[formatter setTimeStyle:NSDateFormatterShortStyle];
+	
+	NSString *dateString = [formatter stringFromDate:[NSDate date]];
+	
+	NSRange amRange = [dateString rangeOfString:[formatter AMSymbol]];
+	NSRange pmRange = [dateString rangeOfString:[formatter PMSymbol]];
+	BOOL is24h = (amRange.location == NSNotFound && pmRange.location == NSNotFound);
+	
+	return is24h;
```
dbart01: What is the point of this method? To check if the current locale uses 24-hour time? It seems extraordinarily wasteful to create a date formatter, render the current date as a string and perform a string search just to identify is a particular locale uses 24-hour time. Correct me if I'm wrong.

ucerxxx: It is not checking locale, it is checking user settings for 24 hour format - I don't know if it is possible to check without it. But I think you are right that it is possible not to do it at all, you can just set some locale (for example "en_GB") insteat or users current and the date object will be ok. To make sure that you understand the problem: set region as United Kingdom and 12 hour format. In this case all dates (coming from your backend) that have HH more than 11 would not be initialized

dbart01: The issue seems to be that the `NSDateFormatter` cannot parse the date string because the format provided by the server doesn't match the one on the client. If that's the case, the solution would be for the server to provide consistent date formats. Do you have an example of the JSON response for the server that breaks using the current implementation?

ucerxxx: Probably it is redundant to change server api for this case. I will explain. Date coming from server as "2017-05-31T14:57:48+01:00" (for example). In this format possible hours value is 00-24. In general dateformatter is working ok with it, but for some locales and 12-hour time mode it can't parse dates with hours more that 11 (example "2017-05-31T11:57:48+01:00" will instantiate proper date, "2017-05-31T14:57:48+01:00" will return nil date object).

Steps to reproduce it on your device:
1. In Settings->General->Language & Region->Region set United Kingdom
2. In time Settings->General->Date & Time set 24-Hour Time off
3. Run the app and check any dates for product in SDK. The dates where HH > 11 will not be instantiated.

Solution. NSDate is in fact Unix timestamp, and "2017-05-31T11:57:48+01:00" includes UTC timezone, so there are no difference what locale you will set to dateFormatter to instantiate NSDate object - it will be the same date. So to avoid #665 issue on affected devices you can just set any locale for NSDateFormatter, instead user's current. 

I did some more investigations of SDK. + (NSDateFormatter*)dateFormatterForPublications is used to instantiate string from date, so users locale should be saved. Check this code, just re-instantiating current user's locale with the same locale solve the problem - dates instantiating normally on affected devices:
NSLocale *currentLocale = [dateFormatter locale];
NSString *identifier = currentLocale.localeIdentifier;

if (![self is24hTimeFormat]) {
	[dateFormatter setLocale:[NSLocale localeWithLocaleIdentifier:identifier]];
}

dbart01: Take a look at this [Stack Overlflow question](https://stackoverflow.com/questions/11062280/nsdateformatter-in-12-hour-mode). Have you tried setting date formatter locale to `en_GB` unconditionally?

ucerxxx: It will work, but I think [dateFormatter setLocale:[NSLocale localeWithLocaleIdentifier:identifier]];

dbart01: Could please remove the `[self is24hTimeFormat]` check (and method) and set the locale explicitly to `en_GB`?

PR #698, opened 2017-07-19
Improved support for checkouts that don't ship

### LINKED ISSUE #696: Crash During Apple Pay Checkout When Shipping Isn't Needed
When attempting to complete an Apple Pay checkout where none of the line items require shipping, the app will crash when attempting to initialize a `PayAddress` with a `PKContact` where the `postalAddress` is nil.

It appears that when the `PKPaymentRequest` is made, if the checkout does not require shipping, a postal address is not requested. https://github.com/Shopify/mobile-buy-sdk-ios/blob/master/Pay/PaySession.swift#L152

However, it is later assumed that there will always be a `postalAddress` on the `PKContact` passed in to `PayAddress`. https://github.com/Shopify/mobile-buy-sdk-ios/blob/master/Pay/PayAddress.swift#L142

Here is a screenshot of the stack trace where the crash happens:
screen shot 2017-07-17 at 4 40 11 pm

dbart01:
### What this does

Adds a new delegate callback for `PaySession` that notifies delegates about shipping address updates. Since billing address isn't available until the end of the Apple Pay flow, we have to require the shipping address even for checkouts that don't require shipping so that we can ask Shopify for relevant tax information to update the subtotal. If we don't do so, the end-user would be authorizing the payment for a subtotal that doesn't include taxes and the amount mismatch would cause the transaction to fail.

Also, allows `PayAddress` to handle a `nil` postal address coming from `PKContact`.

Fixes #696

### DIFF
--- a/Pay/PayAddress.swift
+++ b/Pay/PayAddress.swift
@@ -139,15 +139,25 @@ internal extension PayPostalAddress {
 internal extension PayAddress {
 
     init(with contact: PKContact) {
-        let street = contact.postalAddress!.street
-        let lines  = street.components(separatedBy: .newlines)
+        
+        var line1: String?
+        var line2: String?
+        
+        if let address = contact.postalAddress {
+            let street = address.street
+            let lines  = street.components(separatedBy: .newlines)
+            
+            line1      = lines.count > 0 ? lines[0] : nil
+            line2      = lines.count > 1 ? lines[1] : nil
+        }
+        
         self.init(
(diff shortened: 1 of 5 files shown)


davidmuzi:
fyi: @cocoahero

davidmuzi:
```
❌  /Users/travis/build/Shopify/mobile-buy-sdk-ios/PayTests/Mocks/MockSessionDelegate.swift:30:13: type 'MockSessionDelegate' does not conform to protocol 'PaySessionDelegate'
```

PR #716, opened 2017-08-14
Support for partial postal codes.

dbart01:
### What this does

In Canada and United Kingdom, PassKit vends only partial postal codes. Shopify isn't able to yield shipping rates for partial codes in these countries and so we must pad them on the client. This is done **only** for Canada and United Kingdom. Partial zip codes for United States and other countries are left untouched.

### DIFF
--- a/Pay/PayAddress.swift
+++ b/Pay/PayAddress.swift
@@ -36,25 +36,70 @@ public struct PayPostalAddress {
 
     /// Country (eg: "Canada")
     public let country: String?
+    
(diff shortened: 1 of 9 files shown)


sav007:
I believe we have to provide a good documentation for this. Especially describing the flow how and when to use `zip` vs `paddedZip`

dbart01:
Added documentation, please take a 👀

dbart01:
🎩

screen shot 2017-08-17 at 1 34 54 pm

PR #734, opened 2017-09-07
Input value for explicit nil support

dbart01:
### What this does

Converts all optional fields on input object to type `InputValue<FieldType>` where `FieldType` is the type previously used for those fields. Requiring an `InputValue` forces an explicit decision on whether the values should be set to the following:
- nil
- undefined
- actual value

This change is required to allow update mutation to explicitly send `nil` values in the query, which is particularly useful for nullifying previously non-nil values on the server. Concrete example includes removing a customer's phone number from the server.

#### Philosophy

Other approaches were considered but we ultimately landed on wrapping optional values in a generic `InputValue` type. The intention is to force the user to make an explicit decision when setting values instead of relying on implicit behaviour that isn't documented.

**This is breaking change**. It will be released under a minor version bump - `v3.1`

### DIFF
--- a/Buy/Generated/Storefront/CheckoutAttributesUpdateInput.swift
+++ b/Buy/Generated/Storefront/CheckoutAttributesUpdateInput.swift
@@ -30,15 +30,15 @@ extension Storefront {
 	/// Specifies the fields required to update a checkout's attributes. 
 	open class CheckoutAttributesUpdateInput {
 		/// The text of an optional note that a shop owner can attach to the checkout. 
-		open var note: String?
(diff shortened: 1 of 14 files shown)


davidmuzi:
If this is a breaking change, shouldn't it require a major version bump when using semantic versioning?

davidmuzi:
Worth considering sending empty string when attempting to remove a value, its what our other apps do.

dbart01:
This PR is currently on hold as we work on improving the generator. These changes will also be backwards compatible. Closing for now.

PR #741, opened 2017-09-12
Input for explicit nil support.

dbart01:
### What this does

Converts all optional fields on input object to type `Input<FieldType>` where `FieldType` is the type previously used for those fields. Requiring an `Input` forces an explicit decision on whether the values should be set to the following:
- value(T)
- undefined

This change is required to allow update mutation to explicitly send `nil` values in the query, which is particularly useful for nullifying previously non-nil values on the server. Concrete example includes removing a customer's phone number from the server.

### Other changes from updated generation

- support for `.nodes`
- adds descriptions to various fields

#### Philosophy

Other approaches were considered but we ultimately landed on wrapping optional values in a generic `Input` type. The intention is to force the user to make an explicit decision when setting values instead of relying on implicit behaviour that isn't documented.

**This is non-breaking change**. It deprecates initializers on input objects that contain optional fields and introduces a static `create()` method instead.

### DIFF
--- a/Buy/Generated/Storefront/CheckoutAttributesUpdateInput.swift
+++ b/Buy/Generated/Storefront/CheckoutAttributesUpdateInput.swift
@@ -64,7 +64,7 @@ extension Storefront {
 		///     - customAttributes: A list of extra information that is added to the checkout.
 		///     - allowPartialAddresses: Allows setting partial addresses on a Checkout, skipping the full validation of attributes. The required attributes are city, province, and country. Full validation of the addresses is still done at complete time. 
 		///
-		@available(*, deprecated)
+		@available(*, deprecated, message: "Use the static create() method instead.")
 		public convenience init(note: String? = nil, customAttributes: [AttributeInput]? = nil, allowPartialAddresses: Bool? = nil) {
 			self.init(note: note.orNull, customAttributes: customAttributes.orNull, allowPartialAddresses: allowPartialAddresses.orNull)
 		}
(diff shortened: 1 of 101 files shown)


Buy/Generated/Storefront/Shop.swift
```diff
@@ -297,8 +297,8 @@ extension Storefront {
 		/// - parameters:
 		///     - first: No description
```
davidmuzi: Are these generated? missing `alias` parameter comment

dbart01: Yeah, this is expected behaviour.

Buy/Generated/Storefront/CheckoutAttributesUpdateInput.swift
```diff
@@ -47,25 +47,59 @@ extension Storefront {
...
 		}
 
+		/// Creates the input object.
+		///
+		/// - parameters:
+		///     - note: The text of an optional note that a shop owner can attach to the checkout.
+		///     - customAttributes: A list of extra information that is added to the checkout.
+		///     - allowPartialAddresses: Allows setting partial addresses on a Checkout, skipping the full validation of attributes. The required attributes are city, province, and country. Full validation of the addresses is still done at complete time. 
+		///
+		@available(*, deprecated)
```
sav007: Shouldn't we provide the reason of deprecation, and mention to use `create` instead?

dbart01: Yeah, you're right. Added.

PR #743, opened 2017-09-13
Invalidate cache across version bumps.

### LINKED ISSUE #719: Invalidate cached data with SDK version changes
If a cached response for a query is found, we need to verify that it matches the current SDK version in order to avoid schema collisions.

dbart01:
### What this does

Assigns a framework version to the cache directory. Doing so pins each set of cached queries to the framework version, effectively eliminating any cached data from leaking across releases.

Fixes #719

### DIFF
--- a/Buy/Client/Graph.Cache.swift
+++ b/Buy/Client/Graph.Cache.swift
@@ -32,7 +32,7 @@ internal extension Graph {
     
     internal class Cache {
         
-        static let cacheName = "com.buy.graph.cache"
+        static let cacheName = "com.buy.graph.cache-\(Global.frameworkVersion)"
         
         private static let fileManager = FileManager.default
         


dbart01:
@sav007 If we're using semantic versioning, does it really makes sense to invalidate the cache for _every_ release? Should we only keep track of major releases?

Buy/Client/Graph.Cache.swift
```diff
@@ -32,7 +32,7 @@ internal extension Graph {
     
     internal class Cache {
         
-        static let cacheName = "com.buy.graph.cache"
+        static let cacheName = "com.buy.graph.cache-\(Global.frameworkVersion)"
```
sav007: oh right, I need to do the same on Android

PR #752, opened 2017-09-20
Upgrade to support Xcode 9 and Swift 4.

dbart01:
### What this does

- Upgrades all targets and sample application to Swift 4
- Upgrades `Pay` framework to support new `PassKit` delegates
    - The new delegates are implemented by forwarding calls to existing delegates, which means that `Pay` won't support "custom" error message in the Apple Pay sheet, yet 
- Makes necessary changes to adopt the Swift 4 syntax and naming


pwr7549:
Awesome! How about the Buy tVOS and Buy watchOs?

dbart01:
@pwr7549, take a look at #740 and #749

PR #783, opened 2017-11-10
Require shopName for PaySession

### LINKED ISSUE #781: AppStore rejection because of "PAY TOTAL" string in the ApplePay Sheet
Our app was rejected with such an issue: "Your app uses Apple Pay as a purchasing mechanism but does not use Apple Pay branding and user interface elements appropriately as described in the Apple Pay Human Interface Guidelines. Specifically, the Apple Pay payment sheet does not provide details regarding who the users is paying."

After the investigation of https://developer.apple.com/ios/human-interface-guidelines/technologies/apple-pay/ and a phone conversation with Apple it turned out that thr company/store name is needed to be mention in that line ("PAY TOTAL").

img_3304

As a quick fix to pass the review we've simply changed the string in PayCheckout -> summaryItems: [PKPaymentSummaryItem]
from
```
summaryItems.append(self.paymentDue.summaryItemNamed("TOTAL"))
```
to
```
summaryItems.append(self.paymentDue.summaryItemNamed("<OUR COMPANY>"))
```

After resubmission the app was approved.

Can we expect it to be implemented in future releases of the SDK? 
(It seems like your colleagues on web have already changed it and when using ApplePay from web checkout the sheet looks correct - with a company/store name)

dbart01:
### What this does

Updates `PaySession` to require a `shopName` for creating Apple Pay payment sessions. A shop name is required to accurately reflect the entity to which the payment is being made by the customer.

Updates tests and sample app.

Fixes #781

### DIFF
--- a/PayTests/PaySessionTests.swift
+++ b/PayTests/PaySessionTests.swift
@@ -31,18 +31,28 @@ import PassKit
 @available(iOS 11.0, *)
 class PaySessionTests: XCTestCase {
     
+    private let shopName = "Jaded Labs"
+    
     // ----------------------------------
     //  MARK: - Init -
     //
     func testInit() {
         let checkout = Models.createCheckout()
         let currency = Models.createCurrency()
-        let session  = PaySession(checkout: checkout, currency: currency, merchantID: "some-id")
+        let session  = PaySession(shopName: shopName, checkout: checkout, currency: currency, merchantID: "some-id")
         
         XCTAssertEqual(session.merchantID, "some-id")
         XCTAssertEqual(session.currency.countryCode, currency.countryCode)
         
         XCTAssertFalse(session.identifier.isEmpty)
+    }
+    
+    func testDeprecatedInit() {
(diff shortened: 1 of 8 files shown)


PayTests/PaySessionTests.swift
```diff
@@ -31,20 +31,30 @@ import PassKit
...
         
         XCTAssertFalse(session.identifier.isEmpty)
     }
     
+    func testDeprecatedInit() {
+        let checkout = Models.createCheckout()
+        let currency = Models.createCurrency()
+        let session  = PaySession(checkout: checkout, currency: currency, merchantID: "some-id")
+        
+        XCTAssertEqual(session.shopName, "TOTAL")
```
sav007: Is it valid? Shouldn't it be `Jaded Labs`

sav007: Oh, never mind found

dbart01: Yeah, it the old version.

PR #798, opened 2017-12-20
Don't collapse the Web Checkout's bar on iOS 11

cbowns:
Web Checkout is presented with an SFSafariViewController. We want to disable its bar collapsing, which has to be done at initialization time in its configuration.


cbowns:
Sorry, wrong repo.

PR #803, opened 2018-01-08
Update `update_schema`

### LINKED ISSUE #802: Update introspection endpoint
The current introspection endpoint is deprecated and needs to updated to the new "keyed" endpoint.

dbart01:
### What this does

Update the introspection endpoint in the schema generation script to use the a shared api key.

Fixes #802

### DIFF
--- a/Scripts/update_schema
+++ b/Scripts/update_schema
@@ -7,9 +7,10 @@ require 'optparse'
 require 'graphql_schema'
 require 'graphql_swift_gen'
 
-endpoint = URI("https://app.shopify.com/services/ping/storefront_graphql_schema")
-body     = Net::HTTP.get(endpoint)
-schema   = GraphQLSchema.new(JSON.parse(body))
+shared_storefront_api_key = "e3e948e3697fcc37a837f7dd35719525"
+endpoint                  = URI("https://app.shopify.com/services/graphql/introspection/storefront?api_client_api_key=#{shared_storefront_api_key}")
+body                      = Net::HTTP.get(endpoint)
+schema                    = GraphQLSchema.new(JSON.parse(body))
 
 customScalars = [
 	GraphQLSwiftGen::Scalar.new(
--- a/Scripts/update_schema
+++ b/Scripts/update_schema
@@ -7,7 +7,7 @@ require 'optparse'
 require 'graphql_schema'
 require 'graphql_swift_gen'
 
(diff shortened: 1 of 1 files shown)


Scripts/update_schema
```diff
@@ -7,9 +7,10 @@ require 'optparse'
 require 'graphql_schema'
 require 'graphql_swift_gen'
 
-endpoint = URI("https://app.shopify.com/services/ping/storefront_graphql_schema")
-body     = Net::HTTP.get(endpoint)
-schema   = GraphQLSchema.new(JSON.parse(body))
+shared_storefront_api_key = "e3e948e3697fcc37a837f7dd35719525"
```
chrisbutcher: Just to confirm, do you intend to use the Api Key for https://app.shopify.com/services/internal/api_clients/1424832. It looks like it's the development app, not production? Is there / should there be a a production app?

dbart01: I don't think we've ever had a prod app. @davidmuzi do you have any context on this one?

chrisbutcher: cc: @Shopify/api-patterns-team I think we would always prefer to use a production ApiClient owned by the Shopify partner, even if it's sort of a placeholder (with the correct features, betas switched on) that doesn't actually get installs. Right?

henrytao-me: There is no issue about sharing above key, isn't it?

chrisbutcher: That's right. The ApiClient api key on its own can be considered public.

eapache: This is an SDK not an app; I'm surprised it has a record at all (even for development). Chris, didn't we generate a common baseline key for SDKs to use in this case? It should have no permissions, since we can't guarantee the state of the actual apps using the SDK.

dbart01: Looks like the Unity Buy SDK is using the key from [this app](https://app.shopify.com/services/internal/api_clients/1883616). Should we do the same for iOS and Android?

eapache: I *think* that's the right one? I'll let Chris confirm.

chrisbutcher: Apologies for forgetting this fact. Yes, I can confirm that that is the app we should be using for all Shopify-developed Storefront API apps unless otherwise specified: https://github.com/search?q=org%3AShopify+4a6c829ec3cb12ef9004bf8ed27adf12&type=Code

dbart01: Sweet. I'll make the change.

PR #817, opened 2018-02-28
Update schema.

dbart01:
### What this does

Mostly documentation updates. Add support for WebP image formats via the new `preferredContentType` parameter in `Image`.


davidmuzi:
Should look into https://github.com/github/linguist to have github ignore diffs for generated code

dbart01:
Perhaps, but being mostly a generated SDK, it's oftentimes useful to see these diffs.

PR #926, opened 2018-10-05
Add support for discounts in Apple Pay checkout.

dbart01:
### What this does

| Apple Pay Item Discount | Apple Pay Shipping Discount |
| - | - |
| screen shot 2018-10-05 at 7 04 20 am | screen shot 2018-10-05 at 7 04 06 am |


Sample Apps/Storefront/Storefront/DiscountCodeViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+//  THE SOFTWARE.
+//
+
+import Buy
+
+final class DiscountCodeViewModel: DiscountApplication, ViewModel {
+    
+    typealias ModelType = Storefront.DiscountCodeApplication
+
+    let model: ModelType
```
sav007: I might lost the context but is there any reasons to keep ref to the StoreFront model here? Is it used anywhere?

dbart01: A reference to `ModelType` is a protocol conformance requirement of the `ViewModel` protocol.

PR #974, opened 2019-03-18
Update schema (auto-generated)

rebeccajfriedman:
### What this does

Updates schema (auto-generated) to include:
- Article's SEO title and description
- Bermudian Dollar currency
- Transaction StatusV2 field (which allows null values)


dbart01:
Please use `develop` as the base instead.

PR #991, opened 2019-05-01
Update schema (auto-generated).

dbart01:
### What this does

- Adds support for Product & Variant metafields 🎉


rebeccajfriedman:
Can we hold off on merging this until docs go public (ideally later today)?

PR #1029, opened 2019-09-24
Xcode 11 and iOS 13

dbart01:
### What this does

Update CI to Xcode 11 and iOS 13.


rebeccajfriedman:
Should we wait on merging this until next week when the `2019-10` API release goes live?

Issue #1049, opened 2020-01-20
Multi langauge support [ios-sdk]


osamie:
**Add configurable option to IOS sdk**
- Equivalent change in JS:  https://github.com/Shopify/js-buy-sdk/pull/738/files
- Translation on the storefront API: https://help.shopify.com/en/api/guides/multi-language/translating-content-api#storefront-api-translation-header

osamie:
Code changes: https://github.com/Shopify/mobile-buy-sdk-ios/pull/1050

PR #1050, opened 2020-02-20
Add support for translated fields

osamie:
Fixes https://github.com/Shopify/mobile-buy-sdk-ios/issues/1049

### Add support for translated fields

Storefront API translation header & translatable resources - https://help.shopify.com/en/api/guides/multi-language/translating-content-api#storefront-api-translation-header

**How do merchants use this in their SDK?**
If your store supports multiple languages, Storefront API can return translated resource types and fields. Learn more about [translating content](https://help.shopify.com/en/api/guides/multi-language/translating-content-api).

Initializing the ios sdk to use a specific language
```
 let client: Graph.Client = 
    Graph.Client(
            shopDomain: "your-shop-name.myshopify.com", 
            apiKey: "your-storefront-access-token", 
            locale: Locale.current   // or Locale(identifier:"en-US") 
     )
```

When retrieving the language from the locale, foundation's Locale.identifier gives us a format which is not valid under the spec for use in the `Accept-Language` header. Hence we extend `Locale` to extract the language code and the identifier from the locale in the [correct format for language tag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language).

_Note: readme will be updated in a separate PR_


Buy/Client/Graph.Client.swift
```diff
@@ -79,8 +79,9 @@ extension Graph {
         ///     - shopDomain: The domain of your shop (ex: "shopname.myshopify.com").
         ///     - apiKey:     The API key for you app, obtained from the Shopify admin.
         ///     - session:    A `URLSession` to use for this client. If left blank, a session with a `default` configuration will be created.
+        ///     - locale:    The buyer's locale
         ///
-        public init(shopDomain: String, apiKey: String, session: URLSession = URLSession(configuration: URLSessionConfiguration.default)) {
+        public init(shopDomain: String, apiKey: String, session: URLSession = URLSession(configuration: URLSessionConfiguration.default), locale: Graph.Locale = Graph.Locale.null) {
```
dbart01: Any nested types don't need to be explicitly referenced. We can simplify to this:
```swift
... locale: Locale? = nil) {
```
Side note: Optionality is built into Swift (see comment below). Having said that, I think we should default to `.en` instead of `nil`.

Buy/Client/Graph.Client.swift
```diff
@@ -91,11 +92,13 @@ extension Graph {
                 Header.authorization : apiKey,
                 Header.sdkVersion    : Global.frameworkVersion,
                 Header.sdkVariant    : "ios",
+                Header.sdkLocale     : locale.rawValue,
             ]
 
             precondition(!apiKey.isEmpty, "API Key is required to the Buy SDK. You can obtain one by adding a Mobile App channel here: \(shopURL.appendingPathComponent("admin"))")
         }
 
+
```
dbart01: Extra line.

Buy/Client/Graph.Locale.swift
```diff
@@ -0,0 +1,37 @@
...
+
+import Foundation
+
+public extension Graph {
+    enum Locale: String {
+        case en
+        case fr
+        case de
+        case es
+        case null = ""
```
dbart01: Having a `null` case is redundant. Swift expresses optionality using a `?`. For example:
```swift
let optionalLocale: Locale? = nil
```

Buy/Utilities/Header.swift
```diff
@@ -34,4 +34,5 @@ internal struct Header {
     static var sdkVersion    = "X-SDK-Version"
     static var sdkVariant    = "X-SDK-Variant"
     static var queryTag      = "X-Query-Tag"
+    static var sdkLocale     = "Accept-Language"
```
dbart01: Can we name this after the header? `language` identifies the header more accurately.

Buy/Client/Graph.Client.swift
osamie: Thanks will fix nullable locale param

> Side note: Optionality is built into Swift (see comment below). Having said that, I think we should default to .en instead of nil.

When users of the sdk pass in a value that is not supported, we default not to the en  but the shop langauge

Buy/Client/Graph.Client.swift
```diff
@@ -79,18 +79,20 @@ extension Graph {
...
-            self.apiURL  = Client.urlFor(shopDomain, path: "/api/2019-10/graphql")
+            self.apiURL  = Client.urlFor(shopDomain, path: "/api/2020-01/graphql")
             self.cache   = Cache(shopName: shopDomain)
             self.session = session
             self.headers = [
                 Header.userAgent     : Global.userAgent,
                 Header.authorization : apiKey,
                 Header.sdkVersion    : Global.frameworkVersion,
                 Header.sdkVariant    : "ios",
+                Header.language      : language,
```
dbart01: Using `locale.identifier` will yield a value like `en_US`.

PR #1051, opened 2020-02-25
Locale tests

dbart01:
### What this does

Separates extension on `Locale` into a separate file, adds tests and support for `scriptCode` in multi-component language codes. See tests for details.

### DIFF
--- a/Buy/Utilities/Locale+Language.swift
+++ b/Buy/Utilities/Locale+Language.swift
@@ -1,0 +1,44 @@
+//
+//  Locale+Language.swift
+//  Buy
+//
+//  Created by Shopify.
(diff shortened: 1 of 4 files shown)


Buy/Utilities/Locale+Language.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+
+import Foundation
+
+internal extension Locale {
+    
+    var languageIdentifier: String {
+        let components = [
+            languageCode,
+            scriptCode,
```
osamie: @dbart01 Remove the script code, there are no plans to use the script code at the moment. And the format in the `Accept-Language` header spec does not include it either. Just language code and region/country code.

PR #1056, opened 2020-03-31
Make updating API versioning on the sdk a lot easier  [Part 1]

osamie:
### What am I trying to accomplish?
Make updating the API version on the SDK a lot easier. 
We want to have a single source of truth for the current storefront API version that the SDK is currently using.  

This single source of truth is read by both the code generator and graphql client. 

### Impact
- It becomes possible to automate upgrading of the sdk with the every new storefront API scheduled release.
- It becomes easier for users of the SDK to update the API version of their SDK.

<# Describe your changes #>


Buy/Generated/Storefront.Schema.swift
```diff
@@ -0,0 +1,31 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+extension Storefront {
```
ershkia: can you explain how this is being used?

ershkia: and why hardcoded unlike the other enum?

osamie: @ershkia The "other enum" you are referring to is a template file. The file you're looking at is the output of the generated code, hence its in the folder `Buy/Generated`: 

An example of how this enum is used is found in `Graph.Client.swift` changes

Buy/Client/Graph.Client.swift
```diff
@@ -82,9 +82,9 @@ extension Graph {
         ///     - locale:     The buyer's current locale. Supported values are limited to locales available to your shop.
         ///
         public init(shopDomain: String, apiKey: String, session: URLSession = URLSession(configuration: URLSessionConfiguration.default), locale: Locale? = nil) {
-
+            
             let shopURL  = Client.urlFor(shopDomain)
-            self.apiURL  = Client.urlFor(shopDomain, path: "/api/2020-01/graphql")
+            self.apiURL  = Client.urlFor(shopDomain, path: String(format:"/api/%@/graphql", Storefront.Schema.version))
```
dbart01: The Swift convention is not using `String(format:)`, that's actually an Objective-C function call. With Swift we can composite strings inline:
```swift
Client.urlFor(shopDomain, path: "/api/\(Storefront.Schema.version)/graphql")
```

PR #1057, opened 2020-04-01
Create build.yml

DrTeePot:
### What this does

Auto-runs build script every third month and creates a PR with the result. 

See sample PR here: https://github.com/Shopify/mobile-buy-sdk-ios/pull/1070

Right now it is branching off of develop because that's where the code exists that this needs to work. In the future it could branch off master instead.

Note that once we merge [the release automation](https://github.com/Shopify/mobile-buy-sdk-ios/pull/1065) there will be a check on pull requests to master that makes sure someone has bumped the `podspec` file. 

Also tests will fail if the new generated files aren't added to the project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,72 @@
+name: Build SDK for latest API version
+
+on: 
+  schedule:
+    # Run on midnight UTC on the first of jan/apr/jul/oct
+    # API is released 1PM EDT/EST
+    - cron: '59 23 1 1/3 *'
+
+jobs:
+  build:
+
+    runs-on: macOS-latest
+    
+    steps:
+    - name: Get API Version
+      id: version
+      run: |
+        API_VERSION="$(date +'%Y-%m')"
+        echo "::set-env name=API_VERSION::$API_VERSION"
+      
+    - uses: actions/checkout@master
+      with:
+        ref: 'master'
+      
+    - uses: actions/setup-ruby@v1
+      with:
+        ruby-version: '2.6'
+
+    - name: Install bundler
+      run: gem install bundler
+      
+    - name: Install dependencies
+      run: bundle install
+        
+    - name: Build new SDK version
+      run: |
+        echo "Generating schema for $API_VERSION"
+        ./Scripts/build $API_VERSION
+       
+    - name: Create and reset release branch
+      run: git checkout -B "release/$API_VERSION-auto-generated" # -B will reset branch if it exists
+      
+    - name: Stage and push generated files
+      run: |
+        git add ./Buy/Generated
+        git commit -m "Update schema (auto-generated)"
+        git push --set-upstream --force origin "release/$API_VERSION-auto-generated"
+    
+    - uses: actions/github-script@0.9.0
+      with:
+        script: |
+          // Get the current date
+          const today = new Date();
+          const yyyy = today.getFullYear();
+          const mm = String(today.getMonth() + 1).padStart(2, '0');
+          const api_version = yyyy + '-' + mm
+          
+          // create a pull request
+          github.pulls.create({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'graphql_swift_gen', path: 'Dependencies/Swift Gen'
\ No newline at end of file
+git 'https://github.com/Shopify/graphql_swift_gen' do
+  gem 'graphql_swift_gen', branch: 'task/buy-sdk-3'
```
DrTeePot, line 4: I think we can make this `master`, but this is the branch the sub-module is currently pointed at.

dbart01: `master` has diverged a significant bit from `task/buy-sdk-3`. I'd keep the current branch for the time being.

DrTeePot: Are the tests reliable enough that we could just try changing it (probably not as part of this work) and then see if it matches?

dbart01: No, they're stale, outdated, incomplete and I wouldn't trust them for a second.

dbart01: I did a bunch of work on the SDK branch that isn't in `master`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Create and reset release branch
+      run: git checkout -B test-2020-04 # -B will reset branch if it exists
+      
+    - name: Stage and push generated files
+      run: |
+        git add ./Buy/Generated
+        git commit -m "Test 2020-04"
+        git push --set-upstream origin test-2020-04
+    
+##    - uses: actions/github-script
```
osamie, line 52: Do we still need this? If not remove

DrTeePot: Yes, I'm still working on the tests. This is just commented out because it's improper syntax since it's not done.

.github/workflows/build.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+    - name: Get API Version
+      id: version
+      run: |
+        API_VERSION="$(date +'%Y-%m')"
+        echo "::set-env name=API_VERSION::$API_VERSION"
+      
+    - uses: actions/checkout@master
+      with:
+        ref: 'develop' # This should change to master after 2020-04
```
osamie, line 23: Can we change this to master now and clean up the comment? Or would you rather open another PR to follow this

.github/workflows/build.yml
```diff
@@ -0,0 +1,72 @@
...
+          
+          // create a pull request
+          github.pulls.create({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            title: 'Update schema (auto-generated)',
+            head: api_version + '-auto-generated',
+            base: 'master',
+            body: "Release notes:" + 
+              "\n- https://shopify.dev/concepts/about-apis/versioning/release-notes/" + api_version +
```
osamie, line 66:
```suggestion
              "\n- https://shopify.dev/concepts/about-apis/versioning/release-notes/" + api_version + "#storefront-api-changes"
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,72 @@
...
+              "\n" +
+              "\n**Steps to release:**" +
+              "\n- [ ] Make sure tests are passing" +
+              "\n- [ ] Set the SDK version number (`s.version`) in [`Mobile-Buy-SDK.podspec`](https://github.com/Shopify/mobile-buy-sdk-ios/blob/" + api_version + "-auto-generated/Mobile-Buy-SDK.podspec)" +
+              "\n- [ ] Remember to update the [draft release](https://github.com/Shopify/mobile-buy-sdk-ios/releases) with the same version number"
```
osamie, line 71:
```suggestion
              "\n- [ ] After merging this PR, remember to update the automatically created [draft release](https://github.com/Shopify/mobile-buy-sdk-ios/releases) with the same version number"
```

PR #1058, opened 2020-04-01
Refactored build plus gemfile

DrTeePot:
### What this does

Refactors the script to not assume the location it's being run from. 

Changes gemfile to use git instead of submodules. 

<# Describe your changes #>

Fixes <# issue number #>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'graphql_swift_gen', path: 'Dependencies/Swift Gen'
+git 'https://github.com/Shopify/graphql_swift_gen', branch: 'task/buy-sdk-3' do
+  gem 'graphql_swift_gen'
+end
--- a/Scripts/build
+++ b/Scripts/build
@@ -18,15 +18,20 @@ shared_storefront_api_key = "4a6c829ec3cb12ef9004bf8ed27adf12"
 storefront_api_version = ARGV[0]
 
 endpoint = URI("https://app.shopify.com/services/graphql/introspection/storefront?api_client_api_key=#{shared_storefront_api_key}&api_version=#{storefront_api_version}")
+scripts_path = __dir__
+parent_path = File.expand_path('..', scripts_path)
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'graphql_swift_gen', path: 'Dependencies/Swift Gen'
\ No newline at end of file
+git 'https://github.com/Shopify/graphql_swift_gen', branch: 'task/buy-sdk-3' do
```
osamie, line 3: Neat. What is the advantage of doing this over the git sub module? 
nit: who maintains the `task/buy-sdk-3` branch on the gem's repo?

Scripts/build
```diff
@@ -24,9 +24,9 @@ res = Net::HTTP.get_response(endpoint)
 if res.is_a?(Net::HTTPSuccess)
   puts "Using Storefront API version: #{storefront_api_version}"
 
-  output = ERB.new(File.read("version.swift.erb")).result()
+  output = ERB.new(File.read("./Scripts/version.swift.erb")).result()
```
osamie: So we are assuming that `CWD=mobile-buy-ios/.` everytime the build script is run. However the script itself is in `mobile-buy-ios/Scripts`. 

Have we considered 
- the impact of this change on our current travis job? 

Remember to run it locally

DrTeePot: The previous implementation assumed that this was being run from the scripts directory, and also made assumptions about the computer it was being run on.

Gemfile
DrTeePot: Dima maintains that branch, it's what the current submodule points at. Some versions of bundler don't like relative paths, this was easier than figuring out how to change the github bundler version.

In general I think we should move away from having a git submodule as a dependency, but for now our build and compile also depends on a file in that repo so I can't remove it yet :(

DrTeePot:
@osamie the `task/buy-sdk3` branch is @dbart01's (which the sub-module is currently pinned to) and we can't use `master` since it's diverged.

Deployment docs should still be the same.

PR #1065, opened 2020-04-02
Create release.yml

DrTeePot:
### What this does

Two automations, 
- make sure that the podspec is updated when pushing to master
- release to cocoa and make a draft release when PR is merged into master

Still need to update token in the secrets

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+name: Create SDK release
+
+on: 
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: macOS-latest
+    
+    steps:
+    
+    - uses: actions/checkout@master
+      with:
+        ref: 'master'
+
+    - name: Create a draft release
+      uses: actions/github-script@0.9.0
+      with:
+        script: |
+          // Get the current date
+          const today = new Date();
+          const yyyy = today.getFullYear();
+          const mm = String(today.getMonth() + 1).padStart(2, '0');
+          const api_version = yyyy + '-' + mm
+
+          github.repos.createRelease({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            name: "X.X.X",
+            body: "Release notes:" +
+              "\n- https://shopify.dev/concepts/about-apis/versioning/release-notes/" + api_version + "#storefront-api-changes",
+            tag_name: 'x.x.x',
+            target_commitish: 'master',
+            draft: true,
+            prerelease: false,
+          })
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,30 @@
+name: Deploy to Cocoapods
+
+on:
+  release:
+    types:
+      - published
+      
+jobs:
+  build:
+    runs-on: macOS-latest
+    
+    steps:
+    - uses: actions/checkout@master
+      with:
+        ref: 'master'
+    
+    - uses: actions/setup-ruby@v1
+      with:
+        ruby-version: '2.6' 
+        
+    - name: Install dependencies
+      run: gem install cocoapods
+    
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          const api_version = yyyy + '-' + mm
+
+          github.repos.createRelease({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            name: "ADD VERSION"
+            body: "Release notes:\n - https://shopify.dev/concepts/about-apis/versioning/release-notes/" + api_version,
+            tag_name: 'x.x.x',
+            target_commitish: 'master',
+            draft: true, # TODO this should be false in production
```
osamie, line 42: Who is this TODO for? The only TODO we expect anyone to read should be in the description of the created PR

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          // Get the current date
+          const today = new Date();
+          const yyyy = today.getFullYear();
+          const mm = String(today.getMonth() + 1).padStart(2, '0');
+          const api_version = yyyy + '-' + mm
+
+          github.repos.createRelease({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            name: "ADD VERSION"
```
osamie, line 38: What is the name here? Does this correspond to the title/name of the release? We want the format to be consistent with what we have today

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+    
+    - uses: actions/setup-ruby@v1
+      with:
+        ruby-version: '2.6' 
+        
+    - name: Install dependencies
+      run: gem install cocoapods
+    
+    - name: Create a draft release
+      uses: actions/github-script@0.9.0
```
osamie, line 26: Just curious, could we have used the create-release action for this? https://github.com/actions/create-release#example-workflow---create-a-release

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          const today = new Date();
+          const yyyy = today.getFullYear();
+          const mm = String(today.getMonth() + 1).padStart(2, '0');
+          const api_version = yyyy + '-' + mm
+
+          github.repos.createRelease({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            name: "ADD VERSION"
+            body: "Release notes:\n - https://shopify.dev/concepts/about-apis/versioning/release-notes/" + api_version,
```
osamie, line 39:
```suggestion
            body: "Release notes:\n - https://shopify.dev/concepts/about-apis/versioning/release-notes/" + api_version + "#storefront-api-changes",
```

.github/workflows/release.yml
DrTeePot: Whoever is drafting the release will need to change the name.

.github/workflows/release.yml
DrTeePot: Theoretically yes, but it's much less flexible and people have been having problems with it. Using the script lets us be a lot more explicit about what this action does and will make it obvious what can change if things need to in the future.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+          // Get the current date
+          const today = new Date();
+          const yyyy = today.getFullYear();
+          const mm = String(today.getMonth() + 1).padStart(2, '0');
+          const api_version = yyyy + '-' + mm
+
+          github.repos.createRelease({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            name: "X.X.X"
```
kumaranvram, line 38: missing a comma here. You might encounter an error while running CI. (because we did encounter the same in android)

DrTeePot: Thanks!

PR #1071, opened 2020-04-03
Run tests using Github actions

DrTeePot:
Test new PR's using github actions. This is more transparent IMO than Travis, and we should probably only use one CI tool. 

Theoretically we could run each of these tests in a separate process, but I don't really think it's necessary.


.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: macOS-latest
+
+    steps:
+    - uses: actions/checkout@master
+
+    - uses: actions/setup-ruby@v1
+      with:
+        ruby-version: '2.6'
+    
+    - name: Install xcpretty
```
jmignac, line 21: TIL

.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+      run: gem install xcpretty
+    
+    - name: Update submodules
+      run: git submodule sync --recursive && git submodule update --recursive --init
+      
+    - name: Setup simulator
+      id: version
+      run: |
+        CURRENT_SIMULATOR_UUID=$(xcrun simctl create TestDevice com.apple.CoreSimulator.SimDeviceType.iPhone-11 com.apple.CoreSimulator.SimRuntime.iOS-13-3)
```
jmignac, line 30: I suppose at some point will have to update the simulator reference. Do you see a way of automating that too? Not a blocker

.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
+name: Run tests
+
+on: 
+  pull_request:
+    branches:
```
osamie, line 5: Can you clarify, will these test run if I'm on a different branch trying to merge into master?

DrTeePot: It will run on all pull requests into master or develop

.github/workflows/test.yml
DrTeePot: Not right now. Theoretically you could write a script to check the installed simulators and pick one

DrTeePot: But I think we should consider that an iteration :)

osamie: @jmignac Good question. FYI the changes here in `build.yml` are just a port of how we did it in `travis.yml`. Another infrastructure factor to consider too is that we are running this test on a "prefab environment" `macOS-latest` which is currently catalina 10.15 and may change without our control.    

And yes I agree, this is not a blocker.

.github/workflows/test.yml
osamie: Could we change it to `*` and run on every PR to a branch. I believe that is the current behaviour with travis tests today

PR #1073, opened 2020-04-06
Host sample app project in a separate repository

osamie:
### Summary of change
Removes the sample app from the SDK. 

Add a reference to the new sample app repository - https://github.com/Shopify/mobile-buy-sdk-ios-sample 

Corresponding PR: https://github.com/Shopify/mobile-buy-sdk-ios-sample/pull/1/files


Gemfile
```diff
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'graphql_swift_gen', path: 'Dependencies/Swift Gen'
\ No newline at end of file
+git 'https://github.com/Shopify/graphql_swift_gen', branch: 'task/buy-sdk-3' do
```
DrTeePot, line 3: I'm guessing you're branching off of develop and into master, but maybe should  branch off master in this case. I opened a PR to take care of these commits.

Also I notice you re-based so the Gemfile changes and the build script changes are in one commit, I'd prefer to keep them separate since they were made for different reasons (one fixes a problem with our external dependency management, and one is fixing a scripting problem).

osamie: Done

PR #1074, opened 2020-04-06
Enhancement: Build script

osamie:
### Change summary
Add newly generated source files to the project after build

### DIFF
--- a/Scripts/build
+++ b/Scripts/build
@@ -13,9 +13,11 @@ require 'graphql_schema'
 require 'graphql_swift_gen'
 require 'fileutils'
 require 'erb'
+require 'xcodeproj'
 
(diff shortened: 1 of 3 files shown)


Scripts/build
```diff
@@ -70,8 +77,33 @@ customScalars = [
...
+################################################
+# Add newly generated sources to xcode project
+################################################
+
+project_path = File.join(parent_path, 'Buy.xcodeproj')
+project = Xcodeproj::Project.open(project_path)
+
+generated_files = Dir[File.join(generated_code_path, generated_nesting_class, '*')]
+
+buy_target = project.targets.find{ |t| t.name == 'Buy'}
```
DrTeePot, line 96: This gets the tv and watch targets too?

osamie: @DrTeePot Actually you are right. The generated files are required for all Buy targets. Will update that

PR #1077, opened 2020-04-07
Update Readme

osamie:
### Summary of changes
Reference the new sample app repository in the readme docs.,

### DIFF
--- a/README.md
+++ b/README.md
@@ -1579,7 +1579,8 @@ mutation {
 
 ## Sample application [⤴](#table-of-contents)
 
-The Buy SDK includes a comprehensive sample application that covers the most common use cases of the SDK. It's built on best practices and our recommended `ViewModel` architecture. You can use it as a template, a starting point, or a place to cherrypick components as needed. Check out the [Storefront readme](/Sample%20Apps/Storefront/) for more details.
+For help getting started, take a look at the [sample iOS app](https://github.com/Shopify/mobile-buy-sdk-ios-sample). It covers the most common use cases of the SDK and how to integrate with it. Use the sample app as a template, a starting point, or a place to cherrypick components as needed. Refer to the app's readme for more details.
+
 
 ## Contributions [⤴](#table-of-contents)
 


README.md
```diff
@@ -1579,7 +1579,8 @@ mutation {
 
 ## Sample application [⤴](#table-of-contents)
 
-The Buy SDK includes a comprehensive sample application that covers the most common use cases of the SDK. It's built on best practices and our recommended `ViewModel` architecture. You can use it as a template, a starting point, or a place to cherrypick components as needed. Check out the [Storefront readme](/Sample%20Apps/Storefront/) for more details.
+To help get started, have a look at the [Storefront Sample ios app](https://github.com/Shopify/mobile-buy-sdk-ios-sample). It covers the most common use cases of the SDK and how to integrate with it. You may also use the [Storefront sample android app](https://github.com/Shopify/mobile-buy-sdk-ios-sample) as a template, a starting point, or a place to cherrypick components as needed. Check out the [sample app's readme](https://github.com/Shopify/mobile-buy-sdk-ios-sample) for more details.
```
jmignac, line 1582: Typo: `You may also use the [Storefront sample **android** app](https://github.com/Shopify/mobile-buy-sdk-**ios**-sample)`

DrTeePot: I'm not sure why we reference the android app here at all, would an iOS developer find any value in that?

DrTeePot: I don't think we need to link it three times either

tanderson85: * Following on from Jesse and Tristan's comments, here's a suggested rewording:
    ```suggestion
    For help getting started, take a look at the [Storefront Sample iOS app](https://github.com/Shopify/mobile-buy-sdk-ios-sample). It covers the most common use cases of the SDK and how to integrate with it. Use the sample app as a template, a starting point, or a place to cherrypick components as needed. Refer to the app's readme for more details.
    ```
*What's the actual name of the app as it would appear on the device or App Store? Is it `Storefront App` ? We should render the title accurately and not capitalize any other words. E.g. we might go to `...take a look at **Storefront App**, a sample iOS app.` or something like that. Or if that's too fiddly, we can just refer to `the sample iOS app` and not use the app title

osamie: Thanks @jmignac @DrTeePot @tanderson85 
Tom, I will refer to it as just the sample iOS app. Updated

PR #1079, opened 2020-04-07
Merge develop into master

DrTeePot:
### What this does

Merge develop into master.

We've been merging into master directly, but branching off of develop. Let's just get these out of the way so they stop appearing in every pull request.


DrTeePot:
This already has approvals since the work is from another PR, see #1058  👍

PR #1110, opened 2020-09-23
Swift Package Manager

dbart01:
### What this does

- Updates to Xcode 12 and Swift 5.3
- Removes submodule to `graphql_swift_gen` and integrates all of its source locally (this is required to make Swift Package Manager support work seamlessly)
- Adds `Package.swift` to support SPM along with various updates to sources that we're required as part of this new support
- Updates workflows
  - The `build` workflow now automatically increments the SDK version set internally
  - The incremented version will used in the release workflow to set the correct version of the release

I've noticed that the project versions haven't been updated since `v3.x.x`. These version are important for proper and accurate operation of caching in the SDK as cached queries rely on versioning to invalidate correctly. It's also used in reporting. Every request emitted via the SDK has a header with the SDK version.

The above changes to how versions are handled in workflows will also impact how we query for version internally for caching and reporting. _**It is important that all future updates to the SDK increment this version**_. Doing will be much easier now since most of it will be done automatically and any manual changes will only have to be done once in `SDK.swift`.

### How to review this PR

**TLDR; Please ignore everything under `Dependencies/*`.**

There are a lot of changes. Each major group of changes is broken down into it's own commit. The easiest strategy would be review 1 commit at a time. The largest commit is `Remove submodule` as it adds all Ruby sources from the said submodule.


Buy/Utilities/Global.swift
```diff
@@ -39,7 +39,7 @@ class Global {
     //  MARK: - Versions -
     //
     static var frameworkVersion: String {
-        return self.frameworkBundle.object(forInfoDictionaryKey: "CFBundleShortVersionString") as! String
+        return "\(Storefront.Schema.version)\(SDK.version)"
```
cocoahero, line 42: Should there be some sort of delimiter between these two values? Not sure if it matters for potential server-side reporting.

Scripts/increment_version
```diff
@@ -0,0 +1,22 @@
...
+src = File.read(file_path)
+if !src 
+    puts "Could not read file."
+    exit(false)
+end
+
+version = src.match(/static let version = "([0-9.]+)"/i).captures[0]
+
+components = version.split('.')
+components[1] = components[1].to_i + 1
```
cocoahero, line 14: Are you meaning to be updating the major or minor version here? `X.0.0` or `5.X.0`? I noticed you used `[1]`, but the current history of releases have been major version bumps.

Pay/Utilities/Log.swift
```diff
@@ -0,0 +1,33 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+func Log(_ string: String) {
+#if DEBUG
+    print(string)
```
cocoahero, line 31: nitpick: should we switch to `os_log`?

Buy/Utilities/Global.swift
dbart01: Ah, yes. Good catch. There was supposed to be a `/` in there. Got lost in all the `\`.

cocoahero:
Since the support files were moved out of `/Dependencies`, does this line need to be updated? https://github.com/Shopify/mobile-buy-sdk-ios/blob/spm/Mobile-Buy-SDK.podspec#L28

Scripts/increment_version
dbart01: Yeah, it was meant to update the minor version `0.X.0` because it'd be very unusual for auto-generated changes to be backwards _incompatible_ as that requires changes to the project file. It is possible, though, and it's something that is still left as part of manual process.

Pay/Utilities/Log.swift
dbart01: Ideally, yes. But I'm not sure we have a time budget to tackle that atm.

Buy/Utilities/Global.swift
cocoahero: nitpick: should it be the other way around? `5.1.0/2020-07`?

Buy/Crypto/MD5.swift
```diff
@@ -0,0 +1,111 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import CommonCrypto
+
+enum SHA256 {
```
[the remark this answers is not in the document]
dbart01: What would be the benefit?

Buy/Crypto/MD5.swift
```diff
@@ -0,0 +1,111 @@
+//
+//  SHA256.swift
```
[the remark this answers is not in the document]
dbart01: 🤦  I swear I renamed it.

Buy/Crypto/MD5.swift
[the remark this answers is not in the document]
cocoahero: Probably fair to say it would be premature. I always will suggest using a system framework provided implementation of crypto whenever possible. Maybe we can just wait until we drop iOS 12 and switch entirely and save the polyfill headache.

Buy/Utilities/Global.swift
dbart01: The rationale was that schema version is likely to stay the same _longer_ than SDK version. For example, a subsequent release might "fix bugs" and so the version becomes - `Mobile Buy SDK iOS/2020-07/5.1.1`. Ultimately, I don't have a strong preference here.

cocoahero: Ah, I didn't see the `User-Agent` established format. My opinion would be for using the SemVer approach of `5.1.0+2020-07` as the framework version really is a composite of the two versions, but I am also not strongly held in that regard.

Mobile-Buy-SDK.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
 
   s.name                = 'Mobile-Buy-SDK'
-  s.version             = '5.0.0'
+  s.version             = '5.1.0'
```
cocoahero, line 4: Perhaps this can also be auto-generated in `Scripts/increment_version`?

cocoahero (approved):

PR #1112, opened 2020-09-24
Increment Podspec Version Automatically

cocoahero:
### What This Does
This is a small update to the `increment_version` script that also handles updating the `.podspec`.


dbart01 (approved):
Nice 🚀

PR #1115, opened 2020-09-25
Pin API versions to appropriate months.

dbart01:
### What this does

Obtains the current API version by rounding the month down to the last release instead of using the current month. Doing so let's us manually run the `build` and `release` workflow outside of version release months (at any time).


cocoahero:
- We should probably rename `Scripts/get_version` to `get_sdk_version` to avoid confusion
- Perhaps we can look into adding an argument when the build workflow is invoked manually: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-workflow-configuration

cocoahero (approved):

PR #1249, opened 2024-11-26
Switch to `application/json` request body

cocoahero:
Switches to using `application/json` request body format to comply with the [GraphQL over HTTP spec](https://graphql.github.io/graphql-over-http).

This should be a completely transparent change to existing clients and users.


gmac (approved):
🚀

